PR #5, opened 2016-07-12
Direct link to Error Reporting API

steren:


steren:
https://github.com/GoogleCloudPlatform/cloud-errors-nodejs/pull/9 is about re-writting the quickstart and include this change.

PR #7, opened 2016-07-12
remove local development instructions

steren:


matthewloring:
I don't think this should be removed. These instructions also hold for users hosting their apps in production outside of GCP. Maybe it could be renamed to `Running Elsewhere`?

steren:
Yes I agree the section is not well named, as we say explicitely that the environment should be set to 'production'

I think the README should not assume the user is running on GCP and should contains instructions for running anywhere, and this is what I suggest in https://github.com/GoogleCloudPlatform/cloud-errors-nodejs/pull/9

I suggest to add a new section that details the configuration behavior:
Which environment variables are used and how the various ways to authenticate. What do you think? 

I will do it in this PR if we agree.

ofrobots:
The README.md, in general, needs a fair amount of restructuring to make it flow and sound great. The objective for the current one was to get something basic in place.

I expect there would be some iteration (a.k.a bike-shedding) before it gets to a state that everyone is happy with. @steren: It sound like you're willing to take a stab at driving this overall?

steren:
Yes I can drive having a first version of the README, I will loop you in.

PR #11, opened 2016-07-12
add a .travis.yml

ofrobots:


ofrobots:
Build is broken w/ Node 0.12, but this should be merged regardless.

@GoogleCloudPlatform/node-team PTAL.

.travis.yml
```diff
@@ -0,0 +1,12 @@
+sudo: false
+
+language: node_js
+
+node_js:
+  - "0.12"
+  - "4"
+  - "5"
```
matthewloring: We aren't testing against 5 in our other agents. Is there something special about error reporting that needs 5 integration?

ofrobots:
Preemptively merging. TBR.

.travis.yml
ofrobots: Au contraire: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/.travis.yml

PR #12, opened 2016-07-12
Avoid using Object.assign

ofrobots:
It is not available on Node.js 0.12. Use lodash.assign instead.

@GoogleCloudPlatform/node-team PTAL.


matthewloring:
LGTM.

PR #14, opened 2016-07-12
Add e2e tests

cristiancavalli:
Update documentation and test structure. Update package.json with new integration test command: `npm run-script integration-tests`

PR #17, opened 2016-07-12
Integration tests

cristiancavalli:
Add e2e tests, update docs and test directory structure.


cristiancavalli:
@ofrobots @matthewloring PTAL at this revised version of the PR

ofrobots:
For posterity, this is a continuation of #14.

lib/google-apis/auth-client.js
```diff
@@ -52,14 +59,20 @@ var RetryTracker = require('../classes/retry-tracker.js');
...
 function AuthClient(projectId, shouldReportErrors, key) {
 
   this._authClient = null;
   this._hasInited = false;
+  this._cannotInit = false;
+  this._initError = null;
   this._isRequesting = false;
   this._requestQueue = [];
   this._authFactory = new GoogleAuth();
+  this._eventEmitter = new EventEmitter();
```
ofrobots: You don't need this anymore.

lib/google-apis/auth-client.js
```diff
@@ -52,14 +59,20 @@ var RetryTracker = require('../classes/retry-tracker.js');
  *  service
  * @property {Boolean} _shouldReportErrors - Indicator denoting whether or not
  *  the client should actually attempt requests against the service
+ * @property {EventEmitter} _eventEmitter - The event emitter instance used to
+ *  broadcast the `init` event to listeners so that they can begin queueing
+ *  requests.
```
ofrobots: The JSDoc should mention that this class inherits EventEmitter.

lib/google-apis/auth-client.js
```diff
@@ -69,6 +82,47 @@ function AuthClient(projectId, shouldReportErrors, key) {
...
+inherits(AuthClient, EventEmitter);
+
+/**
+ * Returns the init status of the client. Will return false if the client cannot
+ * init or the client is still waiting to init, will return true if the client
+ * has successfully inited.
+ * @function getInitStatus
+ * @returns {Boolean} - returns the init status of the client
+ */
+AuthClient.prototype.getInitStatus = function() {
```
ofrobots: Does this need to be public?

lib/google-apis/auth-client.js
```diff
@@ -69,6 +82,47 @@ function AuthClient(projectId, shouldReportErrors, key) {
 }
 
 /**
+ * The emit event is emitted when either the underlying transport has
```
ofrobots: s/emit event/init event/

lib/google-apis/auth-client.js
```diff
@@ -83,18 +137,25 @@ AuthClient.prototype._getApplicationDefaultCredentials = function() {
 
     if (err) {
```
ofrobots: style nit: avoid unnecessary blank link after block start, here and elsewhere in the file. This can be fixed in a follow-on (don't pollute this particular PR). Aside: could clang-format be configured to catch this?

lib/google-apis/auth-client.js
```diff
@@ -83,18 +137,25 @@ AuthClient.prototype._getApplicationDefaultCredentials = function() {
...
     if (authClient.createScopedRequired && authClient.createScopedRequired()) {
 
       self._authClient = authClient.createScoped(scopes);
-    } else {
-
-      self._authClient = authClient;
     }
 
     self._hasInited = true;
+    self.emit('init', self.getInitStatus());
```
ofrobots: You're guaranteed that `getInitStatus()` will return `true` here. I think you can get rid of `getInitStatus` unless there is some other need for it that I don't know of.

lib/google-apis/auth-client.js
```diff
@@ -83,18 +137,25 @@ AuthClient.prototype._getApplicationDefaultCredentials = function() {
 
     if (err) {
 
+      console.log('Stackdriver Error Reporting API client cannot initialize:\n',
+                  err, '\n',
+                  'Errors will not be reported to the Stackdriver API.');
+      self._cannotInit = true;
+      self._initError = err;
+      self.emit('init', self.getInitStatus());
```
ofrobots: `getInitStatus` is guaranteed to return `false` here.

lib/google-apis/auth-client.js
```diff
@@ -170,36 +231,83 @@ AuthClient.prototype._makeRequest = function() {
...
- * Checks the length of the `_requestQueue` property on the instance, indicating
- * whether or not another request or set of requests is still pending execution.
+ * Checks the length of the `_requestQueue` property on the instance,
+ * indicating
+ * whether or not another request or set of requests is still pending
+ * execution.
+ * @function _areRequestsStillInQueue
+ * @returns {Number} - the length of the `_requestQueue` property
+ */
+AuthClient.prototype._checkStatusCodeRepeatability =
```
ofrobots: Duplicated from above?

lib/interfaces/hapi.js
```diff
@@ -122,8 +122,8 @@ function hapiRegisterFunction(client, config, server, options, next) {
  */
 function handlerSetup(client, config) {
   var hapiPlugin = {register : hapiRegisterFunction.bind(null, client, config)};
-  var version = isPlainObject(config) && isString(config.version)
-                    ? config.version
+  var version = isPlainObject(config) && isString(config.version) ?
+                    config.version
```
ofrobots: Since you're changing the style; preference would be to keep the ':' operator at the end of current line rather than at the start of the new line. In this case however, the two legs of the conditional are short, so you can both of them on the same line.

tests/integration/testAuthClient.js
```diff
@@ -0,0 +1,394 @@
...
+var lodash = require('lodash');
+var isObject = lodash.isObject;
+var isString = lodash.isString;
+var isEmpty = lodash.isEmpty;
+var has = lodash.has;
+var client;
+
+test(
+  'Test given valid init parameters we should be able to create a client'
+  , function ( t ) {
```
ofrobots: hanging comma. Prefer on the previous line. Doesn't clang-format catch this?

lib/interfaces/hapi.js
cristiancavalli: This is clang formatter automatically formatting this was because of 80col, changing it here would mean a break with other auto-formatted lines

tests/integration/testAuthClient.js
cristiancavalli: @ofrobots It should be, I'll look at it again

lib/google-apis/auth-client.js
cristiancavalli: Fixed

lib/interfaces/hapi.js
ofrobots: I meant that the `config.version` line should be `config.version : '0.0.0';`

Also, for readability, parentheses around `&&` expression.

lib/google-apis/auth-client.js
cristiancavalli: fixed via removal

lib/google-apis/auth-client.js
cristiancavalli: fixed via hardcoded value

lib/google-apis/auth-client.js
cristiancavalli: removed and documentation updated

lib/google-apis/auth-client.js
cristiancavalli: added

lib/google-apis/auth-client.js
cristiancavalli: removed

lib/google-apis/auth-client.js
cristiancavalli: fixed

lib/google-apis/auth-client.js
cristiancavalli: Will follow up with once this has been accepted

lib/google-apis/auth-client.js
```diff
@@ -57,6 +65,8 @@ function AuthClient(projectId, shouldReportErrors, key) {
 
   this._authClient = null;
   this._hasInited = false;
+  this._cannotInit = false;
+  this._initError = null;
```
ofrobots: Could these two properties be elided? The non-nullness of `_initError` implies `_cannotInit` if I am not mistaken.

lib/google-apis/auth-client.js
```diff
@@ -69,6 +79,31 @@ function AuthClient(projectId, shouldReportErrors, key) {
...
+inherits(AuthClient, EventEmitter);
+
+/**
+ * Returns the requesting status of the client, indicating whether or not the
+ * client is currently requesting or not. Will return true if the client is
+ * currently requesting or false if the client is idle.
+ * @function getRequestStatus
+ * @return {Boolean} - the the requesting status of the client
+ */
+AuthClient.prototype.getRequestStatus = function() {
```
ofrobots: This is public only so that tests could call it. It would be preferrable to mark/document this as being for tests only.

lib/google-apis/auth-client.js
```diff
@@ -170,36 +212,61 @@ AuthClient.prototype._makeRequest = function() {
...
+  }
 };
 
 /**
- * Checks the length of the `_requestQueue` property on the instance, indicating
- * whether or not another request or set of requests is still pending execution.
+ * Checks the length of the `_requestQueue` property on the instance,
+ * indicating
+ * whether or not another request or set of requests is still pending
+ * execution.
```
ofrobots: style nit: reflow comment.

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "./node_modules/tape/bin/tape ./tests/unit/*.js",
+    "integration-tests": "./node_modules/tape/bin/tape ./tests/integration/*.js",
```
ofrobots: Should travis run these tests?

tests/integration/testAuthClient.js
```diff
@@ -0,0 +1,394 @@
...
+  }
+);
+
+test(
+  'Given invalid input but a repeatable error response the client should retry'
+  , function ( t ) {
+
+    var tries = 0;
+    var intendedTries = 4;
+    var er = new Error("_@google_STACKWARE_INTEGRATION_TEST_ERROR__");
```
ofrobots: Any significance of the term 'STACKWARE'?

package.json
cristiancavalli: It should, but once it's configured to use an actual account, these integration tests require a valid project

lib/google-apis/auth-client.js
cristiancavalli: fixed

lib/google-apis/auth-client.js
cristiancavalli: Documented

lib/google-apis/auth-client.js
cristiancavalli: merged: a `cannotInit` state is now derived from the null-ness of `initError`

tests/integration/testAuthClient.js
cristiancavalli: fixed

lib/interfaces/hapi.js
cristiancavalli: @ofrobots my bad, this is now fixed

tests/integration/testAuthClient.js
cristiancavalli: My bad changed to `STACKDRIVER`

lib/google-apis/auth-client.js
```diff
@@ -270,11 +358,19 @@ AuthClient.prototype.sendError = function(err, callback) {
 
   if (!this._shouldReportErrors) {
 
-    return;
+    callback(
+        new Error([
+          'Client has not been configured to send errors, please check the',
```
ofrobots: 'Client' would not be descriptive enough to the end user. However, this can be addressed by moving to the diagnostics common library.

tests/integration/testAuthClient.js
```diff
@@ -0,0 +1,394 @@
...
+  function ( t ) {
+
+    if (!isString(process.env.GCLOUD_PROJECT)) {
+      t.fail("The gcloud project id (GCLOUD_PROJECT) was not set as an env variable");
+      t.end();
+      process.exit();
+    }
+
+    try {
+      client = new AuthClient(process.env.GCLOUD_PROJECT, true);
```
ofrobots: I am not familiar with `tape` as a test runner. It seems that these tests have a dependence on order of execution. `client` is a global and this test must run/succeed first for the remainder to be valid. Is there a way to avoid this.

Longer term, I would like all the agents to use the same test library whether it is `tape` or `mocha` or `ava` or something else. /cc @matthewloring for opinions.

ofrobots:
I am okay with this landing now, but I am getting really worried about the duplication and technical debt in the AuthClient w.r.t to diagnostics common library. The request queue specially worries me at this point.

tests/integration/testAuthClient.js
matthewloring: It looks like `tape` doesn't have test lifecycle hooks the way mocha does (https://github.com/substack/tape/issues/59). These tests can be restructured to be order independent (setting the auth client in each test). I don't have a strong preference between test frameworks as we have uniformity between the agents.

ofrobots: >  I don't have a strong preference between test frameworks as we have uniformity between the agents

we don't have uniformity presently.

matthewloring: as long as*

cristiancavalli:
@ofrobots any particular reason the request queue is worrisome? This is functionality we don't currently have in the other client but was asked for in this client

matthewloring:
We should evaluate how soon batching will make it into the errors API. Queuing does not provide much additional value if we are unable to send requests in batch.

ofrobots:
@cristiancavalli I wasn't more specific before because the AuthClient is due for a lot of refactoring, and I am not sure how many of these concerns will remain after the refactoring. 

But given that you asked: the request queue is juggling state across asynchronous continuations and I am not confident (at this point) that there no data races. Furthermore it implements the userspace queuing pattern which will break the trace agent – that is something that will need to be addressed  as well. Given that the API doesn't support batching at this point anyway makes the utility of the queuing even less clear (I might be missing something however).

cristiancavalli:
@ofrobots Understood, thanks for the info. I'm looking at current shared diagnostics client now. One of the questions that keeps popping out at me is how much we would like to build in anticipation of services API's supporting more features vs just having a solution for all three that only meets current needs. I am also a bit worried at the informality of the interface of the current shared diagnostics client. I'm still unclear why it seems to have the ability to cache multiple auth clients, how some request parameters (like retry count) are hardcoded module globals and un-subabble functions like `isTransientError` are determiners of request-retry behavior which seems like potential future pain points for different services.

ofrobots:
> much we would like to build in anticipation of services API's supporting more features

Let's only build what currently works. Unexercised code becomes a bug magnet.

> the ability to cache multiple auth clients

It would make sense to use different auth clients if the request scopes (or other auth parameters) are different; this could be optimized in multiple users are using the same parameters. This smells like premature optimization to me since we are talking N=3 users of the library.

> I am also a bit worried at the informality
> hardcoded module globals and un-subabble functions

Valid concerns. Fixing these is better than not sharing the code.

> I just think it might take some white-boarding

SGTM. Let's do this.

PR #18, opened 2016-07-12
Add coveralls script to package.json

cristiancavalli:
Add support for coveralls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ node_js:
   - "6"
   
 script:
-  - npm run test
+  - npm run-script coveralls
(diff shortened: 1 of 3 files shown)


cristiancavalli:
@ofrobots @matthewloring PTAL

matthewloring:
We will want to get coverage runs going on travis. It can be set up in the travis yaml but it may be easier to emulate the scripts in the trace or debug bin directory.

cristiancavalli:
@matthewloring I just modified the travis.yml file to use the coveralls script which is the same test script just with the output routed to coveralls. This has been tested to work locally when given the correct env variables and the output showed up in coveralls.

ofrobots:
LGTM. 100%?? :dizzy:

PR #22, opened 2016-07-16
Shared diag first integration

cristiancavalli:
### In this iteration:

* Integrate the shared diagnostics library as a dependency
* Replace the AuthClient class with the RequestHandler class
	* A small wrapper around the Diagnostics Lib requester providing a singular interaction interface
* Rewrite AuthClient tests -> RequestHandler tests to maintain 100% test coverage
* Add integration tests to the `coveralls` script now that CI is setup
* Update the travis.yml file to prep for new test runner


cristiancavalli:
@matthewloring @ofrobots PTAL
Since external pr's do not have access to env variables (this means encrypted files as well):
[See here for documentation](https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions)

Then I propose we have a two-legged approach to merging external pr's which I plan to follow for my development flow:
(This PTAL would be the first-leg)

* User identifies a PR they would like to make by opening an issue/contacting team:
	* This is different from the current flow as the user does not pr directly against master
	* The user instead requests a branch is created in the master repo to house the pull request while it is looked over for validity
	* This initial pull request is subject to maintainer critique and unit tests
* Upon acceptance the user pr is merged into the branch, this branch will then me requested to merge against master
	* At this step the merge is subject to acceptance/integration tests since it now has access to the correct env vars
	* If the acceptance tests pass the branch can then be merged into master without further maintainer critique since this has already been accomplished by the first merge, otherwise the pr must be revised and the process start over.

* Reasons:
	* We need acceptance/integration tests to run on every merge into master to make sure we are keeping in lock-step with the API
	* We need a thorough, first-step way of making sure people don't modify security sensitive files like `travis.yml`

ofrobots:
Is it possible to split this PR into two? Logically it seems to be doing two things 1) change how integration tests run on travis and 2) share code from diagnostics-common. Is there a reason why these changes should be made at the same time? I would prefer if travis changes could be made first while not at the same time discussing diagnostic-common.

cristiancavalli:
@ofrobots It's possible but the changes to the travis file are fairly minimal, really just adding an environmental check. The reason I added both is that this commit has 80% to do with adding from shared diag code and the tests around this code can only be activated in the new way that the travis exposes so wouldn't breaking these two up create a pull-request with untestable code?

ofrobots:
Could the travis changes (1) be landed first as a PR on its own and the diagnostic-common work (2) landed as a follow-on? Or are you saying that the (1) is dependent on (2), in which case please elaborate on what the depndence is

cristiancavalli:
@ofrobots The dependency is that landing the pr changes in travis as a first commit would create travis.yml commands which don't reference a file that does actual testing since its in the follow-on pr. The travis changes and the entire pull request are to add the new code and tests. The travis change only adds one new step and that is to accommodate the tests landed as a result of the added shared diag code.

tests/integration/testAuthClient.js
```diff
@@ -46,8 +50,8 @@ test(
     }
    
    t.assert(
-     client instanceof AuthClient,
-     "The client should be an instance of AuthClient"
+     client instanceof RequestHandler,
```
matthewloring: General note that I saw here: Prefer to test interface behavior instead of implementation. Unless you expect the class instance type to be important to end users, I think tests like this will end up introducing maintenance burden.

ofrobots:
Re: workflow, that sounds fine.

I would be also be okay with the following simpler process: 

For maintainer-authored changes:
* maintainer locally runs the integration and e2e tests makes sure they pass.
* maintainer opens a pull request. Travis only runs tests that don't require encrypted data. Landed if green.
* Full testing runs on a merge.

For external PRs the workflow could be:
* external collaborator open pull request. Travis only runs tests that don't require encrypted data.
* maintainer review code, and if it looks good, fetches the code locally. Runs the full tests manually. Grants LGTM if green.
* Full testing runs on a merge.

.travis.yml
```diff
@@ -1,12 +1,22 @@
 sudo: false
-
 language: node_js
-
+before_install:
+- mkdir ./tests/configuration
+- ls ./tests
+- 'if [ "$TRAVIS_TRUSTED_INTERNAL_MERGE" = "true" ]; then openssl aes-256-cbc -K $encrypted_7e22a7c28f69_key -iv $encrypted_7e22a7c28f69_iv -in config.tar.enc -out ./tests/configuration/config.tar -d && tar xvf tests/configuration/config.tar; fi'
```
ofrobots: nit: avoid the temporary file by omitting the `-out` flag to openssl. Use `tar xvf -` to untar from stdin.

lib/google-apis/auth-client.js
```diff
@@ -176,201 +97,68 @@ AuthClient.prototype._generateErrorCreationRequestOptions = function(payload) {
...
-    return;
-  }
-
-  requestPayloadSlot = this._requestQueue[0];
-  thisRequestsPayload = requestPayloadSlot[0];
-  thisRequestsCallback = requestPayloadSlot[1];
-  thisRequestsRetryTracker = requestPayloadSlot[2];
-  thisRequestsOptions =
-      this._generateErrorCreationRequestOptions(thisRequestsPayload);
+RequestHandler.prototype._cbStubNoOp = function (err, response, body) {
```
ofrobots: Does this need to be declared or even exported on the class? Just use an inline no-op function where you need.

lib/google-apis/auth-client.js
```diff
@@ -166,7 +87,7 @@ AuthClient.prototype._generateErrorCreationUrl = function() {
  *  to the service
  * @returns {Object} - The options object for the requesting client
  */
-AuthClient.prototype._generateErrorCreationRequestOptions = function(payload) {
+RequestHandler.prototype._generateErrorCreationRequestOptions = function(payload) {
```
ofrobots: This is rather trivial, and can be inlined into the single caller.

lib/google-apis/auth-client.js
```diff
@@ -15,136 +15,57 @@
...
-  this._initError = null;
-  this._isRequesting = false;
-  this._requestQueue = [];
-  this._authFactory = new GoogleAuth();
-
-  this._key = isString(key) ? key : null;
+function RequestHandler(projectId, shouldReportErrors, key) {
+  
+  this._requestFactory = commonDiag.utils.authorizedRequestFactory([
+    'https://www.googleapis.com/auth/cloud-platform'
```
ofrobots: @steren: the API scope seems really broad. Any reason why error reporting needs full `cloud-platform` scope? Could there be a finer grained scope we could use here?

.travis.yml
```diff
@@ -1,12 +1,22 @@
...
+- export GCLOUD_PROJECT=`cat tests/configuration/stub_project_id.txt`
+- export STUBBED_API_KEY=`cat tests/configuration/stub_api_key.txt`
+- export GOOGLE_APPLICATION_CREDENTIALS=$(pwd)/tests/configuration/stub_default_credentials.json
```
matthewloring: It would be great if we could separate the testing of the auth client (inside diagnostics common) from the testing of the error module. Could the error module behavior be tested with dummy information removing the need for encryption?

cristiancavalli: @matthewloring as we discussed earlier today the only diag code being utilized here is just the bar minimum to get to the point of being able to make an authorized request and the authorization code itself is not directly tested by the error module.

.travis.yml
cristiancavalli: updated

lib/google-apis/auth-client.js
cristiancavalli: changed to inline

lib/google-apis/auth-client.js
cristiancavalli: inlined

.gitignore
```diff
@@ -4,3 +4,7 @@ npm-debug.log
 .DS_Store
 .eslintrc.js
 docs
+stub_api_key.txt
+stub_project_id.txt
+stub_default_credentials.json
+config.tar
```
ofrobots: It would be better to generate these files in a directory and exclude that directory instead of listing all the files individually.

.travis.yml
```diff
@@ -1,12 +1,22 @@
...
-  - "6"
-  
+- '0.12'
+- '4'
+- '5'
+- '6'
 script:
-  - npm run-script coveralls
+- 'if [ "$TRAVIS_TRUSTED_INTERNAL_MERGE" = "true" ]; then npm run-script coveralls; fi'
+- 'if [ "$TRAVIS_TRUSTED_INTERNAL_MERGE" = "" ]; then npm run-script coveralls-external; fi'
```
ofrobots: Could this be an `else` of the above `if` clause?

cristiancavalli: @ofrobots yes but that would make the line extremely long

.gitignore
cristiancavalli: fixed

lib/google-apis/auth-client.js
```diff
@@ -159,218 +80,60 @@ AuthClient.prototype._generateErrorCreationUrl = function() {
...
+      ].join(' ')), null, null);
 
     return;
   }
 
-  this._isRequesting = true;
-  this._makeRequest();
+  this._requestFactory(
+    {
+      url : this._generateErrorCreationUrl(),
```
ofrobots: I think the URL is a constant once the RequestHandler is created. Could you compute and cache this in the constructor? A small function that is only called from the constructor should probably be inlined there.

package.json
```diff
@@ -32,6 +32,7 @@
     "tape": "^4.5.1"
   },
   "dependencies": {
+    "@google/cloud-diagnostics-common": "0.2.3",
     "google-auth-library": "0.9.8",
     "lodash": "4.13.1"
```
ofrobots: Any reason for pinning the version for lodash and google-auth-library? I agree with pinning diagnostics-common.

.travis.yml
ofrobots: http://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines

package.json
cristiancavalli: @ofrobots There has been api breakages with lodash before so I always found pinning on a stable API to be the best, google auth library has been removed as a direct dependency now since we are using the shared diag code

tests/integration/testAuthClient.js
```diff
@@ -59,12 +59,12 @@ test(
...
 
-    client.sendError({}, function (error, result, response) {
+    client.sendError({}, function (error, response, body) {
 
       t.equal(
-        result
+        body,
         null,
-        "The result should be of type null",
+        "The result should be of type null"
```
ofrobots: s/result/body

tests/integration/testAuthClient.js
```diff
@@ -127,6 +128,31 @@ test(
...
 
 test(
+  'Not giving a callback function should still allow the actual request to execute'
+  , function ( t ) {
+
+    var tries = 0;
+    var intendedTries = 5;
+    var er = new Error("_@google_STACKDRIVER_INTEGRATION_TEST_ERROR__");
+    var em = new ErrorMessage()
+      .setMessage(er.stack);
```
ofrobots: nit: fits on the previous line.

tests/integration/testAuthClient.js
```diff
@@ -127,6 +128,31 @@ test(
 );
 
 test(
+  'Not giving a callback function should still allow the actual request to execute'
+  , function ( t ) {
+
```
ofrobots: nit: unnecessary vertical whitespace.

tests/integration/testAuthClient.js
```diff
@@ -127,6 +128,31 @@ test(
...
+    var tries = 0;
+    var intendedTries = 5;
+    var er = new Error("_@google_STACKDRIVER_INTEGRATION_TEST_ERROR__");
+    var em = new ErrorMessage()
+      .setMessage(er.stack);
+    var s = nock(
+      'https://clouderrorreporting.googleapis.com/v1beta1/projects/'+
+        process.env.GCLOUD_PROJECT 
+    ).persist().post('/events:report').reply(200, function () {
+
```
ofrobots: nit: unnecessary vertical whitespace.

tests/integration/testAuthClient.js
```diff
@@ -165,50 +191,46 @@ test(
 );
 
 test(
-  'Launching two requests at the same time should defer the execution of one ' +
-  'request until the other has finished',
+  'Given a key the client should include it as a url param on all requests',
   function (t) {
-
+
```
ofrobots: remove blank line.

tests/integration/testAuthClient.js
```diff
@@ -165,50 +191,46 @@ test(
...
+    
+    var key = process.env.STUBBED_API_KEY
+    client = new RequestHandler(process.env.GCLOUD_PROJECT, true, key);
     var er = new Error("_@google_STACKDRIVER_INTEGRATION_TEST_ERROR__");
     var em = new ErrorMessage()
       .setMessage(er.stack);
+    var s = nock(
+      'https://clouderrorreporting.googleapis.com/v1beta1/projects/' +
+       process.env.GCLOUD_PROJECT 
+    ).persist().post('/events:report?key='+key).reply(200, function (uri) {
```
ofrobots: remove blank line.

tests/integration/testAuthClient.js
```diff
@@ -307,88 +298,30 @@ test(
...
-    process.env.GOOGLE_APPLICATION_CREDENTIALS = "error";
-    client = new AuthClient(process.env.GCLOUD_PROJECT, true);
+    var er = new Error("_@google_STACKDRIVER_INTEGRATION_TEST_API_KEY_ERROR__");
+    var em = new ErrorMessage()
+      .setMessage(er.stack);
+    client = new RequestHandler("_invalid_", true);
 
-    client.on('init',
-      function (initStatus) {
+    client.sendError(em, function (err, response, result) {
```
ofrobots: remove blank line after function start.

tests/integration/testAuthClient.js
```diff
@@ -238,41 +260,10 @@ test(
...
-);
-
-test(
   'Given a configuration to not report errors the client should callback with an error',
   function (t) {
 
-    client = new AuthClient(process.env.GCLOUD_PROJECT, false);
+    client = new RequestHandler(process.env.GCLOUD_PROJECT, false);
 
     client.sendError({}, function (err, result, response){
```
ofrobots: Signature has changed.. should be `err, response, body`.

tests/integration/testAuthClient.js
ofrobots: s/result/body

ofrobots:
Once you have addressed the comments, can you also squash the commits to a minimal number?

ofrobots:
Secondly, I would prefer rebases instead of merges in the commit history.

tests/integration/testAuthClient.js
cristiancavalli: 👍

tests/integration/testAuthClient.js
cristiancavalli: 👍

tests/integration/testAuthClient.js
cristiancavalli: 👍

tests/integration/testAuthClient.js
cristiancavalli: 👍

tests/integration/testAuthClient.js
cristiancavalli: 👍

tests/integration/testAuthClient.js
cristiancavalli: 👍

tests/integration/testAuthClient.js
cristiancavalli: 👍

tests/integration/testAuthClient.js
cristiancavalli: 👍

package.json
```diff
@@ -32,7 +32,7 @@
     "tape": "^4.5.1"
   },
   "dependencies": {
-    "google-auth-library": "0.9.8",
+    "@google/cloud-diagnostics-common": "0.2.3",
     "lodash": "4.13.1"
```
ofrobots: nit: I think it is more important to be able to pickup bug fixes in a timely manner, so it would be better to switch to `^4.13.1` as the dependency for lodash.

ofrobots:
LGTM w/ nit.

cristiancavalli:
pending ci completion

Issue #16, opened 2016-07-12
Change onUncaughtException behaviour


ofrobots:
It would be better to change it to a boolean from an enum.

steren:
This seems to be completed, no?
Please re-open if not.

Issue #29, opened 2016-07-19
I should not need to provide projectId on GAE Flex (use GAE_LONG_APP_ID?)


steren:
On GAE Flex, the GLCOUD_PROJECT environment variable is not available.

I do not expect to need to explicitly provide projectId in my code or setting an environment variable at startup. My GAE instances should be aware of the project then run in.
  
The projectId can be retrieve from the `GAE_LONG_APP_ID`.
As documented [here](https://cloud.google.com/appengine/docs/flexible/nodejs/runtime):

> GAE_LONG_APP_ID is set to the Project ID associated with your application

ofrobots:
I think we should use the Metadata service to query out the project id like we do on the other agents. @steren: would this work for you?

steren:
Sounds good to me, especially if this is what is done in other agents.

steren:
I confirm that by using the code at HEAD, I did not need to provide an explicit 'projectId' to see errors reported in Error Reporting.

PR #30, opened 2016-07-19
New uncaught interface

cristiancavalli:
This addresses ticket #16. This commit builds off pull request #22.


cristiancavalli:
@matthewloring @ofrobots PTAL

ofrobots:
This PR is doing two things at the same time. I would prefer separate commits and/or PRs for distinct things. 
* I want to be able to review the change and test for #16 independently of the change for #29.
* We need a bit of discussion before implementing a fix for #29.
* Commit log: 50 char abstract. We follow the commit guide from node-core: https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#step-3-commit.

cristiancavalli:
@ofrobots pull request changed to focus only on #16

cristiancavalli:
build failed on a coveralls error
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}

restarting..

lib/interfaces/uncaught.js
```diff
@@ -47,14 +46,7 @@ function uncaughtExceptionHandler(client, config, err) {
...
 
-  if (config.onUncaughtException === uncaughtHandlingOptions.report) {
-
-    client.sendError(em);
-  } else if (config.onUncaughtException ===
-             uncaughtHandlingOptions.reportAndExit) {
-
-    client.sendError(em, handleProcessExit);
-  }
+  client.sendError(em, handleProcessExit);
```
ofrobots: You no longer need `config` in this function, so you can remove it from the signature.

lib/configuration.js
```diff
@@ -56,7 +54,7 @@ function extractServiceContext(initConfiguration) {
   };
 
   if (has(initConfiguration, 'serviceContext') &&
-      isObject(initConfiguration.serviceContext)) {
+      isPlainObject(initConfiguration.serviceContext)) {
```
ofrobots: Is this change logically related to how the uncaught exception configuration works?

lib/configuration.js
```diff
@@ -76,9 +74,8 @@ function extractServiceContext(initConfiguration) {
  * @typedef NormalizedConfigurationVariables
  * @type Object
  * @property {String} projectId - the project id of the hosting application
- * @property {uncaughtHandlingEnum}
- *  onUncaughtException=uncaughtHandlingEnum.ignore - the flag denoting how to
- *  handle uncaught exceptions
+ * @property {Boolean} reportUncaughtExceptions=true - the flag denoting
+ *  whether or not to report uncaught errors to the API
```
ofrobots: "whether or not to catch and report uncaught"

lib/configuration.js
```diff
@@ -149,16 +145,20 @@ function gatherConfiguration(initConfiguration) {
         'An invalid configuration was given to the Stackdriver Error Library');
   }
 
+  if (!isString(projectId)) {
+     if (isString(env.GCLOUD_PROJECT) || isNumber(env.GCLOUD_PROJECT)) {
+
+      projectId = env.GCLOUD_PROJECT;
```
ofrobots: This doesn't seem logically related to uncaught exceptions. Can this be moved to a PR of its own?

lib/configuration.js
cristiancavalli: @ofrobots it does not. It was just a quick fix to eliminate a false positive where a user could supply an array.

ofrobots: sgtm

lib/configuration.js
cristiancavalli: fixed

lib/interfaces/uncaught.js
cristiancavalli: fixed

lib/configuration.js
cristiancavalli: @ofrobots sounds good, this change has been reversed

lib/configuration.js
```diff
@@ -149,16 +145,17 @@ function gatherConfiguration(initConfiguration) {
...
-
   if (isString(env.GCLOUD_PROJECT) || isNumber(env.GCLOUD_PROJECT)) {
-
     projectId = env.GCLOUD_PROJECT;
   }
 
+  serviceContext = extractServiceContext(initConfiguration);
+
+
+
```
ofrobots: nit: so much whitespace.

ofrobots:
LGTM w/ nit.

cristiancavalli:
merging after ci pass

Issue #32, opened 2016-07-19
Add travis and david-dm badges


ofrobots:
Because badges!

cristiancavalli:
closing on #33

PR #33, opened 2016-07-19
Add badges

cristiancavalli:
Addresses #32


cristiancavalli:
@ofrobots PTAL

ofrobots:
nit: can we get these on the same line? LGTM otherwise.

cristiancavalli:
@ofrobots updated

cristiancavalli:
Merging now that CI has passed

PR #34, opened 2016-07-20
query project from the metadata service

ofrobots:
Basic support for querying the project from the metadata service. Note
that this is a hacky solution at the moment as we should prefer the
projectId returned from the metadata service in preference to the one
provided by the env. or in the config. This is not done at the moment.

The other problem is that the diagnostics-common module doesn't provide
a mechanism to query the projectId, but numericProjectId instead.
Hopefully numeric id works too.


cristiancavalli:
LGTM, merging

PR #36, opened 2016-07-22
New configuration flow

cristiancavalli:
Add the new Configuration data-sink and add associated
tests and documentation. Update linked interfaces and
update tests for those interfaces.


cristiancavalli:
Build broken by #37

.travis.yml
```diff
@@ -7,6 +7,7 @@ before_install:
 - export GCLOUD_PROJECT=`cat tests/configuration/stub_project_id.txt`
 - export STUBBED_API_KEY=`cat tests/configuration/stub_api_key.txt`
 - export GOOGLE_APPLICATION_CREDENTIALS=$(pwd)/tests/configuration/stub_default_credentials.json
+- alias istanbul="./node_modules/.bin/istanbul"
```
matthewloring: This should be removed, it doesn't appear to solve the problem. I'm not sure how travis/npm run scripts so it may be safer to move the script logic (istanbul/coverage calls) into separate scripts the trace and debug agents are still building and they call istanbul from separate bash scripts.

lib/configuration.js
```diff
@@ -15,150 +15,615 @@
...
+ * @function _assimilateProjectNumber
+ * @param {Error|Undefined} err - If applicable, the error that occured during 
+ *  the transaction
+ * @param {Null|String} projectNum - The project number if successfully 
+ *  retrieved, null otherwise.
+ * @returns {Undefined} - does not return anything
+ */
+Configuration.prototype._assimilateProjectNumber = function(err, projectNum) {
+  if (isNull(err) && isString(projectNum)) {
+    this._projectNumber = projectNum;
```
matthewloring: It might be good to warn here if `_projectNumber` is non-null and doesn't match `projectNum`.

lib/configuration.js
```diff
@@ -15,150 +15,615 @@
...
+  if (this._startedInit === false) {
+    /**
+     * Attempt to get the project number from the metadata service
+     * and kick-off the init process. This would be where we kick off the
+     * `getProjectId` invocation but since that version is not live yet we use
+     * project number for now.
+     * @todo (cristiancavalli) - turn this into the getProjectId call when it
+     *  goes live and remove the getProjectNumber function and call.
+     * @example
+     * utils.getProjectId(this._assimilateProjectId.bind(this));
```
matthewloring: Does assimilateProjectId still exist?

lib/configuration.js
```diff
@@ -15,150 +15,615 @@
...
+ * The _checkLocalServiceContext function is responsible for attempting to
+ * source the _serviceContext objects values from runtime configuration and the
+ * environment. First the function will check the env for known service context
+ * names, if these are not set then it will defer to the _givenConfiguration
+ * property if it is set on the instance. The function will check env variables
+ * `GAE_MODULE_NAME` and `GAE_MODULE_VERSION` for `_serviceContext.service` and
+ * `_serviceContext.version` respectively. If these are not set the 
+ * `_serviceContext` properties will be left at default unless the given runtime
+ * configuration supplies any values as substitutes.
+ * @memberof Configuration
```
matthewloring: Will the @memberof or @function jsdoc annotations ever have a value different from <@memberof>.prototype.<@function> = ... ?

lib/configuration.js
```diff
@@ -15,150 +15,615 @@
...
+ * emitted at which point it will immediately callback to listener.
+ * @memberof Configuration
+ * @public
+ * @function addReadyListener
+ * @chainable
+ * @param {Configuration~readyCallback} callback - the listener to callback on
+ * @returns {this} - returns the instance for chaining
+ */
+Configuration.prototype.addReadyListener = function(callback) {
+  if (this._initState === false && isFunction(callback)) {
```
matthewloring: Prefer `!this._initState`.

lib/configuration.js
```diff
@@ -15,150 +15,615 @@
...
+    // Enforce the one-time fire paradigm
+    this.once('ready', callback);
+  } else if (this._initState === true && isNull(this._initError)) {
```
matthewloring: Prefer `this._initState`.

lib/configuration.js
```diff
@@ -15,150 +15,615 @@
...
+    // Callback immediately as the event already happened
+    callback(this);
```
matthewloring: Users may expect that the callback will always fire asynchronously. If you need to callback immediately it is better to invoke inside a `process.nextTick`.

lib/configuration.js
```diff
@@ -15,150 +15,615 @@
...
-    reportUncaughtExceptions : reportUncaughtExceptions,
-    serviceContext : serviceContext,
-    shouldReportErrorsToAPI : shouldReportErrorsToAPI,
-    key : key,
-    version : version
-  };
-}
-
-module.exports = gatherConfiguration;
+module.exports = Configuration;
```
matthewloring: `\n`

lib/interfaces/hapi.js
```diff
@@ -122,7 +121,7 @@ function hapiRegisterFunction(client, config, server, options, next) {
  */
 function handlerSetup(client, config) {
   var hapiPlugin = {register : hapiRegisterFunction.bind(null, client, config)};
-  var version = (isPlainObject(config) && isString(config.version)) ?
+  var version = isObject(config) && isFunction(config.getVersion) ?
```
matthewloring: Should this be a call to `getVersion`?

tests/integration/testAuthClient.js
```diff
@@ -57,6 +61,7 @@ test(
   'Given invalid json the api should respond with an error message ' +
   'and the callback function should be called',
   function (t) {
+
```
matthewloring: Remove newline.

lib/interfaces/hapi.js
cristiancavalli: Good catch! this is for you 🍰

lib/configuration.js
cristiancavalli: @matthewloring Not that I am aware of, as far as I know, though, this helps with the actual JSDOC generation - grouping the methods under the class page

lib/configuration.js
cristiancavalli: It does not, it was a derivation of the function name `_assimilateProjectNumber`.
I was planning on using that schema `_assimilateProject{id-type}` for the config data sink and `_assimilateProjectId` when the common diag module is published with the new getProjectId functionality

lib/configuration.js
cristiancavalli: @matthewloring That's a good idea, should I land that as a `console.log` in this commit or should I wait till my next PR - I was planning on the next PR swapping out all the logging stuff to the common diagnostics logging stuff

tests/integration/testAuthClient.js
cristiancavalli: :heavy_check_mark:

lib/interfaces/hapi.js
cristiancavalli: ✔️

lib/configuration.js
cristiancavalli: ✔️

lib/configuration.js
cristiancavalli: ✔️

lib/configuration.js
cristiancavalli: ✔️

lib/configuration.js
cristiancavalli: ✔️

lib/configuration.js
cristiancavalli: ✔️

lib/configuration.js
cristiancavalli: ✔️

lib/configuration.js
cristiancavalli: ✔️

lib/interfaces/hapi.js
cristiancavalli: ✔️

cristiancavalli:
@matthewloring @ofrobots Build is now succeeding and green
/play ohyeah

.travis.yml
cristiancavalli: ✔️

lib/configuration.js
```diff
@@ -15,150 +15,614 @@
...
+   * retrieve a project number or a project id from runtime configuration or
+   * from the metadata service then this will trigger the `error` event in which
+   * listening components must operate in 'offline' mode.
+   * {@link https://cloud.google.com/compute/docs/storing-retrieving-metadata}
+   * @memberof Configuration
+   * @private
+   * @type {String|Null}
+   * @defaultvalue null
+   */
+  this._projectNumber = null;
```
ofrobots: Prefer to use ProjectId instead of ProjectNumber.

index.js
```diff
@@ -64,11 +64,8 @@ var uncaughtException = require('./lib/interfaces/uncaught.js');
  */
 function initializeClientAndInterfaces ( initConfiguration ) {
 
-  var config = gatherConfiguration(initConfiguration);
-  var client = new AuthClient(
-    config.projectId,
-    config.shouldReportErrorsToAPI
-  );
+  var config = new Configuration(initConfiguration).init();
```
ofrobots: It seems that the init function is supposed to be called immediately after construction and it must be called. The expectation is that it is called only once. Could the `init` function be merged into the constructor, specially if it ends up simplifying the state machine in `Configuration`: `_initState`, `_startedInit`, and `_initError`.

lib/configuration.js
cristiancavalli: @ofrobots This field actually stores the project number since that is what is released on the diagnostics api right now. When we do a release with the projectId get method then the projectId field will be set

index.js
cristiancavalli: @ofrobots It could be, but then this disallows testing the instances default state when it is initialized and the return values of individual state reduction functions on a yet to be inited instance. I feel that these are valuable tests for determining how a 'un-inited' instance behaves which may occur if there is no user and environmental configuration.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+  this._givenConfiguration = isPlainObject(givenConfig) ? givenConfig : null;
+  /**
+   * The _logger property holds a logger instance which, if valid, will be used
+   * to log any debug or error information generated by the instance.
+   * @memberof Configuration
+   * @private
+   * @type {Object|Null}
+   * @defaultvalue null
+   */
+  this._logger = isObject(logger) ? logger : null;
```
ofrobots: Prefer to throw is logger is null. Assume logger is non-null in the rest of the file

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
-    if (has(initConfiguration, 'projectId')) {
-
-      projectId = initConfiguration.projectId;
-    }
-
-    if (has(initConfiguration, 'reportUncaughtExceptions')) {
-
-      reportUncaughtExceptions = !!initConfiguration.reportUncaughtExceptions;
+    if (this._logger) {
+      this._logger.warn(
```
ofrobots: This shouldn't necessarily warn. The metadata service is only accessible on GCP. When running locally, it is expected that we will be unable to retrieve the projectId automatically. Here's the logic we use on trace: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/2167b16941b89a33aed6d9e3bebc3f9e9cb534d0/index.js#L153

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+        'Unable to retrieve project number\n',
+        '\t', 'Project Number:', projectNum, '\n',
+        '\t', 'Error:', err
```
ofrobots: Prefer to provide actionable error message. What can the user do to rectify the situation. The trace agent example is a good one.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+ * @public
+ * @function getProjectNumber
+ * @returns {Number|Null} - returns the _projectNumber property
+ */
+Configuration.prototype.getProjectNumber = function(cb) {
+  var self = this;
+  if (!isFunction(cb)) {
+    throw new Error('Must be supplied with a callback');
+  } else if (!isNull(self._projectNumber)) {
+    process.nextTick(function () {
```
ofrobots: Prefer `setImmediate` instead: https://gist.github.com/brycebaril/ff86eeb90b53fd0c523e.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+      cb(self._projectNumber);
+    }.bind(this));
```
ofrobots: Why bind?

lib/configuration.js
ofrobots: Okay this ends up working because the trace agent seems to be using local config first before checking with the metadata service. (/cc @matthewloring to reconcile the ordering). Regardless the warning should only show up when the user has done something wrong.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+ * number (e.g. '123'). If this environmental variable is not set the function
+ * will defer to the _givenConfiguration property if it is of type object and
+ * has a string property named projectId. If none of these conditions are met
+ * then the _projectNumber property will be left at its default value.
+ * @memberof Configuration
+ * @private
+ * @function _checkLocalProjectNumber
+ * @returns {Undefined} - does not return anything
+ */
+Configuration.prototype._checkLocalProjectNumber = function() {
```
ofrobots: Do we need both `_checkLocalProjectNumber` and `_checkLocalProjectId`?

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+   * The _givenConfiguration property holds a ConfigurationOptions object
+   * which, if valid, will be merged against by the values taken from the meta-
+   * data service. If the _givenConfiguration property is not valid then only
+   * metadata values will be used in the Configuration instance.
+   * @memberof Configuration
+   * @private
+   * @type {Object|Null}
+   * @defaultvalue null
+   */
+  this._givenConfiguration = isPlainObject(givenConfig) ? givenConfig : null;
```
ofrobots: Initialize `_givenConfiguration` to `{}` instead of `null` and get rid of isPlainObject checks from the rest of the file.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
-      isPlainObject(initConfiguration.serviceContext)) {
-
-    if (has(initConfiguration.serviceContext, 'service')) {
-
-      defaultReturnObject.service = initConfiguration.serviceContext.service;
+  var gaeModuleVer = env.GAE_MODULE_VERSION;
+  this._serviceContext.service = isString(gaeModuleName) ? gaeModuleName : '';
+  this._serviceContext.version = isString(gaeModuleVer) ? gaeModuleVer : '';
+  if (isPlainObject(this._givenConfiguration)
+    && isPlainObject(this._givenConfiguration.serviceContext)) {
```
ofrobots: `&&` hangs off the previous line

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+ * configuration supplied values but also the processes environmental values.
+ * @memberof Configuration
+ * @private
+ * @function _gatherLocalConfiguration
+ * @returns {Undefined} - does not return anything
  */
-
+Configuration.prototype._gatherLocalConfiguration = function() {
+  if (isPlainObject(this._givenConfiguration)) {
+    if (isBoolean(this._givenConfiguration.reportUncaughtExceptions)) {
```
ofrobots: Should we silently ignore things uncaught exceptions if the user accidentally sets this to '1' instead of `true`? Either throw or be more resilient to bad user input and do the right thing.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+      this._reportUncaughtExceptions = this._givenConfiguration
+        .reportUncaughtExceptions;
+    }
+    if (isString(this._givenConfiguration.key)) {
```
ofrobots: Throw instead of ignoring when the key isn't a string.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+    throw new Error('Must be supplied with a callback');
+  } else if (!isNull(this._projectId)) {
+    process.nextTick(function () {
+      cb(self._projectId);
+    });
+  } 
+  // @TODO {cristiancavalli} this is where the utils project async call will go 
+  else {
+    
+    process.nextTick(function () {
```
ofrobots: `setImmediate`.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+ * @public
+ * @function getProjectId
+ * @returns {String|Null} - returns the _projectId property
+ */
+Configuration.prototype.getProjectId = function(cb) {
+  var self = this;
+  if (!isFunction(cb)) {
+    throw new Error('Must be supplied with a callback');
+  } else if (!isNull(this._projectId)) {
+    process.nextTick(function () {
```
ofrobots: `setImmediate`.

lib/google-apis/auth-client.js
```diff
@@ -78,6 +75,59 @@ function getErrorReportURL(projectId, key) {
...
+ *  for communication to the StackDriver API
+ * @param {Function} cb - the function to callback on once the request has
+ *  completed/errored
+ * @param {String} projectIdentifier - String uniquely identifying the target
+ *  project
+ * @returns {Undefined} - does not return anything
+ */
+RequestHandler.prototype._doRequest = function (errorMessage, cb, projectIdentifier) {
+  if (isNull(projectIdentifier)) {
+    process.nextTick(function () {
```
ofrobots: `setImmediate`.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+   * the `NODE_ENV` environmental variable. If the `NODE_ENV` variable is set to
+   * 'production' then the _shouldReportErrorToAPI property will be set to true
+   * and error reporting library will attempt to send errors to the Error API.
+   * Otherwise the value will remain false and errors will not be reported to
+   * the API.
+   * @memberof Configuration
+   * @private
+   * @type {Boolean}
+   */
+  this._shouldReportErrorsToAPI = env.NODE_ENV === 'production';
```
ofrobots: A better name for this property would be `_enabled`.

lib/google-apis/auth-client.js
```diff
@@ -92,35 +142,13 @@ RequestHandler.prototype.sendError = function(errorMessage, userCb) {
   var that = this;
   var cb = isFunction(userCb) ? userCb : function() {};
 
-  if (!that._shouldReportErrors) {
-
+  if (that._config.getShouldReportErrorsToAPI()) {
+    this._getProjectIdentifier(this._doRequest.bind(this, errorMessage, cb));
```
ofrobots: Bound functions are significantly slower (until recently). `_doRequest` doesn't seem to have other uses. Is there a strong reason to change this from the original that was saving state through closure capture instead of `this` & bind?

lib/interfaces/manual.js
```diff
@@ -83,7 +81,7 @@ function reportManualError(client, config, err, request, additionalMessage,
 
     em.setMessage(am);
   }
-
+
```
ofrobots: spurious whitespace?

lib/interfaces/express.js
```diff
@@ -46,10 +47,10 @@ function expressErrorHandler(client, config, err, req, res, next) {
   var ctxService = '';
   var ctxVersion = '';
 
-  if (isPlainObject(config) && isPlainObject(config.serviceContext)) {
+  if (isObject(config) && isFunction(config.getServiceContext)) {
```
matthewloring: In what situation do we expect `getServiceContext` to not be a function?

cristiancavalli: duly noted redundancy will remove

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+  this._key = null;
+  /**
+   * The _serviceContext property is meant to contain the optional service
+   * context information which may be given in the runtime configuration. If
+   * not given in the runtime configuration then the property value will remain
+   * null.
+   * @memberof Configuration
+   * @private
+   * @type {Object}
+   * @default
```
matthewloring: What is the `@default` tag?

cristiancavalli: @matthewloring supposed to be `@defaultvalue {Object}` (its null right now but that will change with the update to this pr today). Thanks!

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+ * @function _checkLocalProjectId
+ * @returns {Undefined} - does not return anything
+ */
+Configuration.prototype._checkLocalProjectId = function() {
+  if (isString(this._projectId)) {
+    // already has been set by the metadata service
+    // this call is still waiting to go live in the utils library
+    return;
+  } else if (isString(env.GCLOUD_PROJECT) && !isEmpty(env.GCLOUD_PROJECT)
+    && isNaN(env.GCLOUD_PROJECT)) {
```
ofrobots: Keep `&&` on the previous line.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+    // GCLOUD_PROJECT is set and it is not the project number, set on instance
+    this._projectId = env.GCLOUD_PROJECT;
+    return;
+  } else if (isPlainObject(this._givenConfiguration)
+    && !isEmpty(this._givenConfiguration.projectId)
+    && isString(this._givenConfiguration.projectId)
+    && isNaN(this._givenConfiguration.projectId)) {
```
ofrobots: ditto x 3.

lib/configuration.js
matthewloring: We should decide on whether we intend to generate jsdoc at some point. Thus far we've used annotations strictly to improve code readability but I do see the value if we generate jsdoc. I think this decision will help determine which annotations will be helpful. cc @ofrobots

lib/interfaces/express.js
matthewloring: I think more generally, typechecks of this form can be replaced with unit tests. Validating input crossing the boundary from user code into our module makes sense but I think typechecking our own internal representations is better validated by unit tests to allow us to catch issues during development as opposed to at runtime.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
-  var reportUncaughtExceptions = true;
-  var serviceContext = null;
-  var key = null;
-  var shouldReportErrorsToAPI = false;
-
-  if (env.NODE_ENV === 'production') {
-
-    shouldReportErrorsToAPI = true;
+Configuration.prototype._assimilateProjectNumber = function(err, projectNum) {
+  if (isNull(err) && isString(projectNum)) {
```
matthewloring: I would generally prefer `!err` to `isNull(err)`. `isNull(undefined)` will evaluate to `false` causing us to ignore the provided project number when an `err` of `undefined` should indicate no issue. Also, the param annotation above would indicate that `err` can only be `Null` and not `undefined`.

lib/configuration.js
```diff
@@ -15,150 +15,430 @@
...
+ * The _gatherLocalConfiguration function is responsible for determining
+ * directly determing whether the properties `reportUncaughtExceptions` and
+ * `key`, which can be optionally supplied in the runtime configuration, should
+ * be merged into the instance. This function also calls several specialized
+ * environmental variable checkers which not only check for the optional runtime
+ * configuration supplied values but also the processes environmental values.
+ * @memberof Configuration
+ * @private
+ * @function _gatherLocalConfiguration
+ * @returns {Undefined} - does not return anything
```
matthewloring: Do we need to specify return annotations on functions without return values?

cristiancavalli: I prefer it as a matter of style - always be explicit about return values:
https://github.com/eslint/eslint/issues/673

lib/configuration.js
cristiancavalli: I disagree with this since `!""` is true. Given that the underlying library explicitly returns null in the error function when success occurs whereas an empty string, given that the underlying interface has already changed which we should be aware of since we administer a wrapper over it, could be a perfectly valid error, just one that has no content.

lib/configuration.js
cristiancavalli: I think _enabled is too general. What is enabled? The class? A property of the class? Makes it seem like the class itself is enabled when it actually has nothing to do with the class itself - the class is instead simply a container for another explicit configuration

lib/configuration.js
ofrobots: I don't see JSDoc generation as a high priority issue as this would be internal documentation for a module that is unlikely to grow beyond a small size.

lib/interfaces/hapi.js
matthewloring: Can we add the test that would have caught this?

lib/configuration.js
matthewloring: It is correct that we use the project id discovered from the environment or config before we go to the metadata service. Is that what you intended?

cristiancavalli: The flow of usage is not dictated here. The getProjectId function is called from the constructor because the method to get the project id from the service is not yet on the npm repo and can only be gathered locally at the moment. The auth client still attempts to get the project number first since that can be gathered from the metadata service using the existing api

lib/configuration.js
cristiancavalli: @ofrobots  removing this since this is better fixed by the follow-on commit which introduces the logger a require

lib/configuration.js
cristiancavalli: @ofrobots for right now yes since they are checking different things

cristiancavalli:
@ofrobots @matthewloring now new and improved

lib/configuration.js
```diff
@@ -15,150 +15,429 @@
...
- * @property {String} service - the hosting applications underlying service
- *  platform
- * @property {String} version - the hosting applications version
+ * The Configuration constructor function initializes several internal
+ * properties on the Configuration instance and accepts a runtime-given
+ * configuration object which may be used by the Configuration instance
+ * depending on the initialization transaction that occurs with the meta-data
+ * service.
+ * @class Configuration
+ * @extends EventEmitter
```
ofrobots: This no longer needs needs to be an EventEmitter, right? The following classdesc needs to be updated as well.

lib/configuration.js
ofrobots: diagnostics 0.2.4 is now released. You can remove all code related to projectNumber from here.

lib/configuration.js
```diff
@@ -15,150 +15,429 @@
...
+ * @returns {Undefined} - does not return anything
  */
-
+Configuration.prototype._gatherLocalConfiguration = function() {
+  if (isBoolean(this._givenConfiguration.reportUncaughtExceptions)) {
+    this._reportUncaughtExceptions = this._givenConfiguration
+      .reportUncaughtExceptions;
+  } else if (has(this._givenConfiguration, 'reportUncaughtExceptions'))  {
+    throw new Error([
+      'If specifying the configuration option reportUncaughtExceptions a',
```
ofrobots: Simplify to `config.reportUncaughtExceptions must be a boolean`.

lib/configuration.js
```diff
@@ -15,150 +15,429 @@
...
+ * environmental variable checkers which not only check for the optional runtime
+ * configuration supplied values but also the processes environmental values.
+ * @memberof Configuration
+ * @private
+ * @function _gatherLocalConfiguration
+ * @returns {Undefined} - does not return anything
  */
-
+Configuration.prototype._gatherLocalConfiguration = function() {
+  if (isBoolean(this._givenConfiguration.reportUncaughtExceptions)) {
```
ofrobots: style nit: reverse the condition and throw. This makes this clause unconditional, which is much easier to read.

lib/configuration.js
```diff
@@ -15,150 +15,429 @@
...
+    this._reportUncaughtExceptions = this._givenConfiguration
+      .reportUncaughtExceptions;
+  } else if (has(this._givenConfiguration, 'reportUncaughtExceptions'))  {
+    throw new Error([
+      'If specifying the configuration option reportUncaughtExceptions a',
+      'boolean must be supplied.'
+    ].join(' '));
+  }
+  if (isString(this._givenConfiguration.key)) {
```
ofrobots: likewise.

lib/configuration.js
```diff
@@ -15,150 +15,429 @@
...
+    this._key = this._givenConfiguration.key;
+  } else if (has(this._givenConfiguration, 'key')) {
+    throw new Error([
+      'If specifying the configuration option key a',
```
ofrobots: `config.key must be a string`.

lib/configuration.js
```diff
@@ -15,150 +15,429 @@
...
+ * @function _checkLocalProjectId
+ * @returns {Undefined} - does not return anything
+ */
+Configuration.prototype._checkLocalProjectId = function() {
+  if (isString(this._projectId)) {
+    // already has been set by the metadata service
+    // this call is still waiting to go live in the utils library
+    return;
+  } else if (isString(env.GCLOUD_PROJECT) && !isEmpty(env.GCLOUD_PROJECT) &&
+    isNaN(env.GCLOUD_PROJECT)) {
```
ofrobots: Are you missing a '!' before `isNan`? I think we might be missing some tests if this got missed.

lib/google-apis/auth-client.js
```diff
@@ -89,38 +109,30 @@ function getErrorReportURL(projectId, key) {
...
-  var that = this;
+  var self = this;
   var cb = isFunction(userCb) ? userCb : function() {};
-
-  if (!that._shouldReportErrors) {
-
+  if (self._config.getShouldReportErrorsToAPI()) {
+    self._getProjectIdentifier(function (projectIdentifier) {
+      if (isNull(projectIdentifier)) {
+        process.nextTick(function () {
```
ofrobots: Use `setImmediate`.

lib/configuration.js
cristiancavalli: @ofrobots No this is check to see if it is not a number otherwise, if it is a number, it is considered to be a project id

lib/configuration.js
cristiancavalli: @ofrobots this would make the statement `!isBoolean && has` which is much longer than the current if/else. It's perfectly valid that this property in the configuration is not a boolean, its just that the key shouldn't have been inited on the object. If it was and its not a boolean then it is not valid.

lib/configuration.js
matthewloring: My general understanding of the contract of an "errorback" style function is that if the error value is truthy then an error occurred and the other parameters should not be considered. Otherwise, the error value will be falsey to indicate that no error occured. The specific falsey or truthy values passed to the first argument are implementation details and a typecheck like this will start failing without violating the errorback contract if a different falsey value is specified. Since `''` is a falsey value, I would consider this as no error occurring and a bug on the calling end if it should be treated as such.

lib/configuration.js
cristiancavalli: likewise to the complexity expansion of the above statement. There are two paths that are valid: when the user sets a key with a correct type in the config & when leaves the field empty and that value is substituted from the env. There is one invalid case: when a user submits a config key with an invalid type - this switch demonstrates both branches and only tests one condition on each

tests/integration/testAuthClient.js
```diff
@@ -253,54 +258,109 @@ test(
...
+        'The status code should be 200'
+      );
+      process.env.GCLOUD_PROJECT = oldProject;
+      process.env.STUBBED_API_KEY = oldKey;
+      t.end();
+    });
   }
 );
+
+  // test(
```
matthewloring: Should this be removed or commented back in?

lib/configuration.js
cristiancavalli: @matthewloring Point well-taken. Although it makes a bit uncomfortable to use 'truthiness' here I see that its a pattern in the node community. Will update accordingly in pr update.

cristiancavalli:
@ofrobots @matthewloring ready for another look

lib/configuration.js
ofrobots: Sure the condition is longer, but the logic is much more readable:

```js
Configuration.prototype._gatherLocalConfiguration = function() {
  if (has(this._givenConfiguration, 'reportUncaughtExceptions') &&
      !isBoolean(this._givenConfiguration.reportUncaughtExceptions)) {
    throw new Error('config.reportUncaughtExceptions must be a boolean');
  }
  if (has(this._givenConfiguration, 'reportUncaughtExceptions' &&
      !isString(this._givenConfiguration.key) {
    throw new Error('config.key must be a string');  
  }

  this._reportUncaughtExceptions = this._givenConfiguration
      .reportUncaughtExceptions;
  this._key = this._givenConfiguration.key;
};
```

lib/configuration.js
```diff
@@ -15,150 +15,335 @@
...
+ * property named projectId. If none of these conditions are met then the
+ * _projectId property will be left at its default value.
+ * @memberof Configuration
+ * @private
+ * @function _checkLocalProjectId
+ * @param {Function} cb - The original user callback to invoke with the project
+ *  id or error encountered during id capture
+ * @returns {Undefined} - does not return anything
+ */
+Configuration.prototype._checkLocalProjectId = function(cb) {
```
ofrobots: This function seems to be sync. Why does it take a callback instead of returning?

lib/configuration.js
```diff
@@ -15,150 +15,335 @@
...
-  var projectId = null;
-  var reportUncaughtExceptions = true;
-  var serviceContext = null;
-  var key = null;
-  var shouldReportErrorsToAPI = false;
-
-  if (env.NODE_ENV === 'production') {
-
-    shouldReportErrorsToAPI = true;
+Configuration.prototype._assimilateProjectId = function(err, projectId, cb) {
```
ofrobots: This function is sync, why does it accept a callback instead of returning the answer? This also looks like a candidate for inlining into the single caller since this is so simple and the term 'assimilate' doesn't intrinsically describe what this function does.

lib/configuration.js
cristiancavalli: @ofrobots because getProjectId is async and having this internal function respond to the callback makes it easier to separate out is specific logic versus the public getter function.

lib/configuration.js
cristiancavalli: ack

tests/test-servers/express_scaffold_server.js
```diff
@@ -67,7 +63,7 @@ app.get(
 
           console.log(EXCLAMATION_LN);
           console.log("Successfully sent custom get error to api");
-          console.log(res);
+          // console.log(res);
```
matthewloring: Uncomment or delete.

tests/test-servers/express_scaffold_server.js
```diff
@@ -113,7 +109,7 @@ function reportManualError ( ) {
 
         console.log(EXCLAMATION_LN);
         console.log("Successfully sent error information to the API");
-        console.log(res);
+        // console.log(res);
```
matthewloring: Uncomment or delete.

lib/configuration.js
```diff
@@ -15,150 +15,316 @@
...
+/**
+ * Returns the _projectId property on the instance.
+ * @memberof Configuration
+ * @public
+ * @function getProjectId
+ * @returns {String|Null} - returns the _projectId property
+ */
+Configuration.prototype.getProjectId = function(cb) {
+  var self = this;
+  if (!isFunction(cb)) {
```
matthewloring: Is this a function we expect end users to call or can this validation be eliminated?

matthewloring:
Looks good on my end w/ nits.

lib/configuration.js
cristiancavalli: @matthewloring @ofrobots This is an semi-philosophical question to me:
Should we maintain interfaces between our components which break quickly when used incorrectly as a way of 'defining' the interface or should they fail silently? I prefer the former since as codebases grow in size and more people begin to work on them I find that interfaces which break when used incorrectly are easier to develop with and expect certain behaviour from.

lib/configuration.js
cristiancavalli: @ofrobots ack, this is better. Will update accordingly

lib/configuration.js
matthewloring: I completely agree that well-defined interfaces are beneficial to a growing project. Without this check, this function will still throw an exception if no callback is provided. Clean error messages make a lot of sense when they will be surfaced to users but "future proofing" error handling that will never be exercised in the current code base can easily cause unnecessary clutter.

cristiancavalli: Point well taken, I think in this case the default stack-trace that will be provided should be sufficient, will update accordingly.

lib/configuration.js
cristiancavalli: @ofrobots as per our last conversation keeping this the same due to the valid configuration but specific key is not set case.

cristiancavalli:
@ofrobots @matthewloring Feedback integrated

ofrobots:
LGTM.

Issue #26, opened 2016-07-18
Don't require invocation on import


JustinBeckwith:
Currently we require you to require like this:
```js
var errors = require('@google/cloud-errors')();
```

May developers (myself included) are going to miss the `()` required at the end before using things like the hapi error handler.  Is there any way to hang the methods and objects off of the exported module by default?

PR #43, opened 2016-07-31
Add common-diagnostics logger

### LINKED ISSUE #15: Use @google/cloud-diagnostics-common
Some of the functionality (logger, auth, retry) should be refactored/reused from the @google/cloud-diagnostics-common module.

### LINKED ISSUE #40: Warning should be logged if error cannot be reported
If an error cannot be reported (the API returns something else than 200), a warning should be logged saying that the error could not be reported.

As far as I know, it is not the case today: I have an app that reports errors, errors are not appearing in the UI despite the message payload being a valid stacktrace.

I would guess the API returns something else than 200 if called.

cristiancavalli:
07/30/2016-20:46 PST
Add the common diag logger to the library and replace all console.log
calls with the common diag logger equivalent calls. Add default logging
of errors on init/transaction errors as per #40 when user does not
supply a callback or is using a server plugin. Add tests to address new
behaviour.
Fixes #15 
Fixes #40 

08/12/2016-23:33 PST
Add the Common Diagnostics Logger to the error library and create
logging points for failures in both configuration and interaction with
the Stackdriver Error Reporting API.

### DIFF
--- a/lib/google-apis/auth-client.js
+++ b/lib/google-apis/auth-client.js
@@ -16,6 +16,7 @@
 
 'use strict';
 var commonDiag = require('@google/cloud-diagnostics-common');
+var logger = require('../logger.js');
 var lodash = require('lodash');
 var isFunction = lodash.isFunction;
 var isString = lodash.isString;
@@ -66,7 +67,7 @@ function RequestHandler(config) {
  * @private
  */
 function getErrorReportURL(projectId, key) {
-  var url = [ API, projectId, 'events:report'].join('/');
+  var url = [API, projectId, 'events:report'].join('/');
   if (isString(key)) {
     url += '?key=' + key;
   }
@@ -91,18 +92,36 @@ RequestHandler.prototype.sendError = function(errorMessage, userCb) {
     self._config.getProjectId(function (err, id) {
       if (err) {
         setImmediate(function () { cb(err, null, null); });
+        logger.error([
+          'Unable to retrieve a project id from the Google Metadata Service or',
+          'the local environment. Client will not be able to communicate with',
+          'the Stackdriver Error Reporting API without a valid project id', 
+          'Please make sure to supply a project id either throught the'
+          'GCLOUD_PROJECT environmental variable or through the configuration',
+          'object given to this library on startup if not running on Google',
+          'Cloud Platform.'
+        ].join(' '));
         return;
       }
       self._request({
         url: getErrorReportURL(id, self._config.getKey()),
         method: 'POST',
         json: errorMessage
-      }, cb);
+      }, function (err, response, body) {
+        if (err) {
+          logger.error([
+            'Encountered an error while attempting to transmit an error to the',
+            'Stackdriver Error Reporting API.'
+          ].join(' '), err);
+        }
+        cb(err, response, body);
+      });
     });
   } else {
     cb(new Error([
-      'Client has not been configured to send errors, please check the',
-      'NODE_ENV environment variable and make sure it is set to production'
+      'Stackdriver error reporting client has not been configured to send',
+      'errors, please check the NODE_ENV environment variable and make sure it',
+      'is set to production'
     ].join(' ')), null, null);
   }
 };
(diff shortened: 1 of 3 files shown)


cristiancavalli:
@matthewloring @ofrobots PTAL

matthewloring:
This appears to have config changes, test script changes, and diagnostic common changes. Could these be separated into different PRs for review?

cristiancavalli:
@matthewloring The actual commit here is dc80657, the other commits underneath it are the other pr #36. This commit will build off of what that commit merges into master

JustinBeckwith:
I am a simple man :) Could you include doc changes (readme) with any API changes?  It makes seeing the public facing API changes much easier for me.

cristiancavalli:
@JustinBeckwith Sorry if I mislead you, this change doesn't introduce any API changes; #44 does though and that fixes #26

JustinBeckwith:
Apologies I misread the changelog.  Thanks!

lib/logger.js
```diff
@@ -0,0 +1,3 @@
+var logger = require('@google/cloud-diagnostics-common').logger;
+var logLevel = process.env.GCLOUD_DEBUG_LOGLEVEL;
+module.exports = logger.create(logLevel, '@google/cloud-errors');
```
matthewloring: '\n'

lib/logger.js
```diff
@@ -0,0 +1,3 @@
+var logger = require('@google/cloud-diagnostics-common').logger;
+var logLevel = process.env.GCLOUD_DEBUG_LOGLEVEL;
```
matthewloring: I think we will want our own env variable for this rather than tying ourselves to the debug log level.

cristiancavalli:
@matthewloring @ofrobots PTAL

lib/logger.js
```diff
@@ -0,0 +1,3 @@
+var logger = require('@google/cloud-diagnostics-common').logger;
+var logLevel = process.env.GCLOUD_ERRORS_LOGLEVEL;
+module.exports = logger.create(logLevel, '@google/cloud-errors');
```
matthewloring: '\n'

lib/logger.js
```diff
@@ -0,0 +1,3 @@
+var logger = require('@google/cloud-diagnostics-common').logger;
+var logLevel = process.env.GCLOUD_ERRORS_LOGLEVEL;
```
matthewloring: It would be good to support setting this through the configuration object as well. It is fine to open a bug and do that in a follow on.

cristiancavalli: Will add in later follow on

cristiancavalli:
Updated

lib/logger.js
matthewloring: Opened here: https://github.com/GoogleCloudPlatform/cloud-errors-nodejs/issues/51

lib/google-apis/auth-client.js
```diff
@@ -91,18 +92,34 @@ RequestHandler.prototype.sendError = function(errorMessage, userCb) {
     self._config.getProjectId(function (err, id) {
       if (err) {
         setImmediate(function () { cb(err, null, null); });
+        logger.error([
+          'Unable to retrieve a project id from the Google Metadata Service or',
+          'the local environment. Client will be unable to communicate with',
```
ofrobots: nit: `s/will be unable/will not be able/`

lib/google-apis/auth-client.js
```diff
@@ -91,18 +92,34 @@ RequestHandler.prototype.sendError = function(errorMessage, userCb) {
...
+          'the Stackdriver Error Reporting API without a valid project id', 
+          'Please make sure to supply a GCLOUD_PROJECT environmental variable',
```
ofrobots: Either through the env. var. or in the config object.

lib/logger.js
ofrobots: As written, I think we disable all logging if the env. var. is not set. This would be a major regression. This needs to be fixed before #51 (or #51 needs to be merged back into this issue).

cristiancavalli: @ofrobots @matthewloring The default behavior of the logger is not log given that it the log level is undefined:
https://github.com/GoogleCloudPlatform/cloud-diagnostics-common-nodejs/blob/master/lib/logger.js#L48

Therefore this behavior should match the description right?

ofrobots: I think we should log at the warn level if the environment variable is not set.

cristiancavalli:
@ofrobots @matthewloring Updated

ofrobots:
lgtm.

PR #44, opened 2016-08-01
Remove invocation on import and create configure property

### LINKED ISSUE #26: Don't require invocation on import

cristiancavalli:
Remove the anonymous function invocation required for init and replace
with a plain JS object exported with a start method which calls the
otherwise anonymously invoked function. Update documentation and
relevant tests.

Fixes #26

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,8 @@ applications running in almost any environment. Here's an introductory video:
 
 	```JS
 	// Require the library and initialize the error handler
-	var errors = require('@google/cloud-errors')({
-		serviceContext: {service: 'my-service'}	// not needed on Google App Engine
+	var errors = require('@google/cloud-errors').start({
+		serviceContext: {service: 'my-service'} // not needed on Google App Engine
 	});
 
 	// Report an error to the Stackdriver Error Reporting API
@@ -61,7 +61,7 @@ When initing the Stackdriver Error Reporting library you must specify the follow
 On Google App Engine, these environment variables are already set.
 
 ```JS
-var errors = require('@google/cloud-errors')({
+var errors = require('@google/cloud-errors').start({
 	projectId: 'my-project-id',
 	key: 'my-api-key',
 	keyFilename: 'path-to-my-keyfile'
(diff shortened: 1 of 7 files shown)


cristiancavalli:
@matthewloring @ofrobots PTAL

README.md
```diff
@@ -35,8 +35,8 @@ applications running in almost any environment. Here's an introductory video:
 
 	```JS
 	// Require the library and initialize the error handler
-	var errorHandler = require('@google/cloud-errors')({
-		serviceContext: {service: 'my-service'}	// not needed on Google App Engine
+	var errorHandler = require('@google/cloud-errors').configure({
```
JustinBeckwith: I thought we were doing .start() for consistency with debug/trace?  @ofrobots

ofrobots: Yes, `.start` is what we discussed on chat.

README.md
```diff
@@ -78,7 +78,8 @@ var errorHandler = require('@google/cloud-errors')({
 ```JS
 var express = require('express');
 var app = express();
-var errorHandler = require('@google/cloud-errors')();
+// Will create a errorHandler instance based off env variables
+var errorHandler = require('@google/cloud-errors');
```
ofrobots: I don't think we should be highlighting any API other than `.start`. I think it would be *nice* that we would help/warn the user if they forgot to call `start`. I would like to keep the API surface area as minimal as possible.

cristiancavalli:
@ofrobots @matthewloring @JustinBeckwith PTAL

README.md
```diff
@@ -35,8 +35,8 @@ applications running in almost any environment. Here's an introductory video:
 
 	```JS
 	// Require the library and initialize the error handler
-	var errors = require('@google/cloud-errors')({
-		serviceContext: {service: 'my-service'}	// not needed on Google App Engine
+	var errorHandler = require('@google/cloud-errors').start({
```
ofrobots: I think you're accidentally renaming `errors` back to `errorHandler` as part of a merge conflict resolution (in general, you can avoid these by pulling `master` frequently). Can you put these back?

ofrobots:
LGTM w/ comment.

PR #46, opened 2016-08-02
Avoid using bind in various handlers

ofrobots:
* Switched various handlers to use closures to save state instead of using bind.
* Simplify comments to avoid verbalizing information already available in the type annotations
* restify.js is pending similar treatment


lib/interfaces/hapi.js
```diff
@@ -53,75 +53,65 @@ function hapiErrorHandler(req, err, config) {
...
- * the errors can be logged to the Error Reporting API.
- * @function hapiRegisterFunction
- * @param {AuthClient} client - A bound reference to the auth client for
- *  communication with the Google Error API
+ * Creates a Hapi plugin object which can be used to handle errors in Hapi.
+ * @param {AuthClient} client - an inited auth client instance
  * @param {NormalizedConfigurationVariables} config - the environmental
  *  configuration
- * @param {Hapi.Server} server - A Hapi server instance
- * @param {Object} options - The server configuration options object
```
cristiancavalli: @ofrobots Why are you removing these comments over the interfaces?

ofrobots: GitHub is failing at showing the diff. These are being moved, not removed.

cristiancavalli: Oh I see, my bad - thanks for the clarification

lib/interfaces/express.js
```diff
@@ -25,66 +25,56 @@ var expressRequestInformationExtractor =
...
- * stack into the Express architecture. This function accepts four arguments:
- * a bound reference to a client which should be given by the proceeding
- * handlerSetup function, an error argument which can be of any type, an
- * express request object, an express response object and the express next
- * function for passing input up the plugin chain.
- * @function expressErrorHandler
- * @param {AuthClient} client - a bound and inited Auth Client instance
- * @param {NormalizedConfigurationVariables} config - the environmental
+ * Returns a function that can be used as an express error handling middleware.
+ * @function makeExpressHandler
```
matthewloring: While we're in here, is `@function` necessary?

lib/interfaces/express.js
```diff
@@ -25,66 +25,56 @@ var expressRequestInformationExtractor =
...
-function expressErrorHandler(client, config, err, req, res, next) {
-  var ctxService = '';
-  var ctxVersion = '';
+function makeExpressHandler(client, config) {
 
-  if (isPlainObject(config) && isPlainObject(config.serviceContext)) {
+  /**
+   * The Express Error Handler function is an interface for the error handler
+   * stack into the Express architecture.
+   * @function expressErrorHandler
```
matthewloring: And here?

ofrobots: ack.

lib/interfaces/express.js
ofrobots: ack. will remove.

cristiancavalli: @matthewloring The rest of the library uses this format, not necessary but I prefer to make the tags explicit to avoid run-on with the preamble

lib/interfaces/uncaught.js
```diff
@@ -86,13 +38,30 @@ function attachUncaughtExceptionHandler(client, config) {
...
+    client.sendError(em, handleProcessExit);
+  }
 
   if (config.reportUncaughtExceptions === false) {
     // Do not attach a listener to the process
     return null;
   }
 
-  return attachUncaughtExceptionHandler(client, config);
+  return process.on('uncaughtException', uncaughtExceptionHandler);
```
matthewloring: Does this return the `process` object?

lib/google-apis/auth-client.js
```diff
@@ -24,7 +24,7 @@ var isString = lodash.isString;
 var SCOPES = ['https://www.googleapis.com/auth/cloud-platform'];
 
 /* @const {String} Base Error Reporting API */
-var API = 'https://clouderrorreporting.googleapis.com/v1beta1/projects'
+var API = 'https://clouderrorreporting.googleapis.com/v1beta1/projects';
```
matthewloring: Is `jshint` running on each change? I would have thought it caught things like this

lib/interfaces/express.js
ofrobots: What do you mean by 'run-on with the preamble'. My preference is DRY.

lib/interfaces/manual.js
```diff
@@ -26,68 +26,7 @@ var manualRequestInformationExtractor =
...
-/**
- * The interface for manually reporting errors to the Google Error API in
- * application code, this interface takes five arguments, three of which can
- * be supplied by application-code invocation and only one of which is required.
- * The first arugment `client` will be bound to an instance of this function so
- * that this handler can send error information back to Error Reporting API. The
- * second argument `err` is the actual error information that the application is
- * attempting to report. This data can be of any type, but better errors will be
- * logged given a valid instance of Error/Exception class, this is the only
- * required argument to be supplied during invocation. The third argument
```
cristiancavalli: @ofrobots Why is this preamble removed in the new function? This describes the behavior of the function and the new preamble is now a seemingly redundant description of what the name of the function already describes

lib/google-apis/auth-client.js
ofrobots: It is not. I want to fix this as a follow-on, but holding back to avoid a merge conflict with a currently open PR that also touches package.json.

lib/interfaces/express.js
```diff
@@ -25,66 +25,56 @@ var expressRequestInformationExtractor =
...
+   * @returns {ErrorMessage} - Returns the ErrorMessage instance
+   */
+  function expressErrorHandler(err, req, res, next) {
+    var ctxService = '';
+    var ctxVersion = '';
 
-    ctxService = config.serviceContext.service;
-    ctxVersion = config.serviceContext.version;
-  }
+    if (isPlainObject(config) && isPlainObject(config.serviceContext)) {
```
matthewloring: What does this provide over `config && config.serviceContext`? I appreciate this is not related to the refactor at hand but I was curious.

lib/interfaces/express.js
cristiancavalli: It's sometimes hard to tell the difference sometimes between the preamble finish and the name of the function. This is also an accepted and valid style in JSDOC: http://usejsdoc.org/tags-function.html

lib/interfaces/express.js
cristiancavalli: isPlainObject checks for whether or not the given value is not function or array. This simply narrows the margin for error in attempting to access values out of a data structure that won't yield those results

lib/interfaces/manual.js
ofrobots: This description talks in detail about the number of arguments, which ones are optional, and what their purpose is. I have moved such details to the @-param documentation where it made sense.

lib/interfaces/uncaught.js
cristiancavalli: Yes, the `.on` function returns the EventEmitter instance

lib/interfaces/express.js
matthewloring: My understanding was that the `@function` tag was used primarily to denote variables of function type. If I had:

```javascript
function g() {
  return function() {};
}
/** @function */
var i = g();
```

It may be difficult to determine that `i` held a function without looking up the definition of `g`. 

I think having an `@function` tag on a function declaration such as:

```javascript
/** @function */
function f() {};
```

may be unnecessary.

lib/interfaces/uncaught.js
matthewloring: Why does it return the `process` object?

lib/interfaces/express.js
cristiancavalli: Sure its optional in this context, but its also just making it explicit to the parser. There are plenty of examples of functions using the [at]function tag during declaration, it seems like a matter of style

lib/interfaces/uncaught.js
```diff
@@ -86,13 +38,30 @@ function attachUncaughtExceptionHandler(client, config) {
...
+  /**
+   * The actual exception handler creates a new instance of `ErrorMessage`,
+   * extracts infomation from the propagated `Error` and marshals it into the
+   * `ErrorMessage` instance, attempts to send this `ErrorMessage` instance to
+   * the Stackdriver Error Reporting API. Subsequently the process is
+   * terminated.
+   * @function uncaughtExceptionHandler
+   * @listens module:process~event:uncaughtException
+   * @param {Error} err - The error that has been uncaught to this point
+   * @returns {Undefined} - does not return a value
```
cristiancavalli: If you're returning the process object this should probably be updated

lib/interfaces/express.js
matthewloring: In general, I would prefer validating an interface as opposed to an implementation. If we were to change the representation of `config` to a function or class that had the appropriate properties still defined this check would begin failing without modifying the interface of the object. If we do begin passing `config` incorrectly or with the wrong values it would be better for unit tests to catch the error than for runtime checks to silence it.

cristiancavalli: @matthewloring this change is already pending: https://github.com/GoogleCloudPlatform/cloud-errors-nodejs/pull/36/files#diff-cfd0a7cde5f1b09ea6953801a995b14e

matthewloring: Ok, I can raise the question there

matthewloring:
LGTM.

lib/interfaces/express.js
ofrobots: Google style guide doesn't require [at]function, but it allows for it: https://google.github.io/styleguide/javascriptguide.xml?showone=Comments#Comments. 

With that style guide given, I think we should favour DRY. The guideline Matt pointed to makes sense to me.

cristiancavalli:
LGTM

lib/interfaces/uncaught.js
ofrobots: Not returning anything from this function.

lib/interfaces/uncaught.js
ofrobots: `process` object is an `EventEmitter` instance. Is your question about why we are returning anything at all? It is a common pattern to chain `EventEmitter.on` calls. From that point of view, it makes sense to return the `EventEmitter` (i.e. `process`).

PR #47, opened 2016-08-02
Add tests and fix arguments for manual report

ofrobots:
Positional optional arguments required a bit rigidity.


tests/unit/testManualHandler.js
```diff
@@ -0,0 +1,136 @@
...
+test('undefined additional message', function(t) {
+  var r = report('Mickey', { method: 'SNIFF'}, undefined, function(err, res) {
+    t.end();
+  });
+  t.equal(r.message, 'Mickey', 'string error should propagate');
+  t.equal(r.context.httpRequest.method, 'SNIFF');
+});
+
+test('arguments after the callback funciton should get ignored', function(t) {
+  var r = report('honkey', function(err, res) {
```
cristiancavalli: @ofrobots Maybe a different term? This is a bit controversial

cristiancavalli:
LGTM - but could the test cases include the if that coveralls found? Seems like an good branch to test

tests/unit/testManualHandler.js
ofrobots: Thanks for catching this! I have changed it. 😨

cristiancavalli:
Looks great!

ofrobots:
Fixed a test to properly document check a specific edge case behaviour. Please take another look at the latest commit.

cristiancavalli:
@ofrobots thanks for edge-case documentation - this LGTM

PR #49, opened 2016-08-13
Remove keyfile option from configuration

cristiancavalli:
Remove keyfile option from configuration because we do not support a keyfile through the runtime configuration since we are using the util library and the google-auth-library does not accept a key file name in its public interface.


cristiancavalli:
@matthewloring @ofrobots PTAL

ofrobots:
The interface is not the best, but I think you can use `.fromStream` from googleauth. https://github.com/google/google-auth-library-nodejs/blob/229b0ca66b48ee32292c55b922a12aba3232aa00/test/test.googleauth.js#L271

/cc @tbetbetbe

ofrobots:
or alternatively `.fromJSON`.

@tbetbetbe: is there a reason that the `google-auth-library` doesn't accept a filename and instead requires the library users to do a JSON parse / provide a read stream?

ofrobots:
This change LGTM since a proper fix is dependant on upstream fix: https://github.com/google/google-auth-library-nodejs/pull/92.

PR #50, opened 2016-08-13
Add timer to uncaught exception reporter for process exit

### LINKED ISSUE #35: Uncaught exception handler should terminate with a timeout
```JS
function uncaughtExceptionHandler(client, config, err) {
  var em = new ErrorMessage();

  errorHandlerRouter(err, em);

  client.sendError(em, handleProcessExit);
}
```

If things are completely b0rked, `sendError` might never succeed, in which case process may never exit. I think we should follow a policy similar to the trace agent where we expose a delay in the config after which we terminate the process regardless. WDYT?
/cc @matthewloring

cristiancavalli:
Update uncaught handler to terminate on timeout #35
    
Add a setTimeout call for 2000ms to exit after a uncaught exception has
been caught when `config.reportUncaughtExceptions` is set to true. Add
tests and behaviour fixture to test full control flow from when handler
initialization to error report to process close.
    
Related issues:
Fixes #35

### DIFF
--- a/lib/interfaces/uncaught.js
+++ b/lib/interfaces/uncaught.js
@@ -54,6 +54,7 @@ function handlerSetup(client, config) {
 
     errorHandlerRouter(err, em);
     client.sendError(em, handleProcessExit);
+    setTimeout(handleProcessExit, 2000);
   }
 
   if (config.reportUncaughtExceptions === false) {
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -38,6 +38,7 @@ function decryptAcceptanceCredentials {
 function runFullSuite {
   $(npm bin)/istanbul cover -x "fuzzer.js" \
   $(npm bin)/tape ./tests/integration/*.js ./tests/unit/*.js \
+  ./tests/fixtures/*.js \
   --report lcovonly -- -R spec && cat ./coverage/lcov.info \
   | $(npm bin)/coveralls && rm -rf ./coverage
 }
--- a/tests/fixtures/uncaughtExitBehaviour.js
+++ b/tests/fixtures/uncaughtExitBehaviour.js
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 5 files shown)


cristiancavalli:
@matthewloring @ofrobots PTAL

lib/interfaces/uncaught.js
```diff
@@ -54,6 +54,7 @@ function handlerSetup(client, config) {
 
     errorHandlerRouter(err, em);
     client.sendError(em, handleProcessExit);
+    setTimeout(handleProcessExit, 2000);
```
matthewloring: Could we add a test to ensure that errors are published when an uncaught exception happens (if not tested elsewhere)?

cristiancavalli: @matthewloring  This would be difficult to do/gain value out of since we already test the transport layer and sending messages seperatley while to test this we would need to either rely on a timer and attempt to check for a new error (inaccurate and flaky since we can't query for specific errors) or attempt to monkey path the library itself for testing so that we can do the same integration test we already to with the transport layer with the only difference being the call from the uncaught exception event listener. Essentially these components are individually tested and trying to insert instrumentation to test the two working together would require changing the way the components work so that we would be able to test them.

matthewloring: Sorry, publish may have been a misleading word. I want to make sure that we have a test that fails if we cease to report errors if an unhandled exception occurs. Here is what we do in the trace agent: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/test/standalone/test-trace-uncaught-exception.js. There are some timers in play but we have not had flakiness troubles with this test on the trace end.

cristiancavalli: @matthewloring I added two types of test for this case:
1) A unit test file to cover API surface
2) A spawn-able test fixture which runs tests against the lifecycle of this uncaught exception behavior, including: grabbing the exception data, reporting it to the (mocked) endpoint and terminating the fork in under 2500ms.

matthewloring: Looks great!

matthewloring:
LGTM when tests are green.

PR #53, opened 2016-08-22
Update defaults for service context on Error Message #52

### LINKED ISSUE #52: Defaults for serviceContext
If not available and not provided, the library should use clever default for serviceContext:

- leave the serviceContext.version to not set (the UIs will probably handle this better than empty string)
- set serviceContext.service to 'node' (this is better than empty string or 'default')

cristiancavalli:
Update the defaults on the Error Message `serviceContext` object
property to API-accepted defaults for `service` and `version`
subproperties. Set `serviceContext.service` to default to `'node'` and
`serviceContext.version` to `undefined` so that if the property remains
'unset' it will be excluded from the JSON rendering for API OTW
consumption. Fix syntax for error message production in
`auth-client.js`.

Original defaults:

```JS
{serviceContext: {service: '', version: ''}}
```

Updated defaults:

```JS
{serviceContext: {service: 'node', version: undefined}}
```

Related issues:
Fixes #52

### DIFF
--- a/lib/classes/error-message.js
+++ b/lib/classes/error-message.js
@@ -43,8 +43,8 @@ var isObject = lodash.isObject;
  * @property {Object} serviceContext - The service information for the error
  * @property {String} serviceContext.service - The service that the error was
  *  was produced on
- * @property {String} serviceContext.version - The service version that the
- *  error was produced on
+ * @property {String|Undefined} serviceContext.version - The service version 
+ *  that the error was produced on
  * @property {String} message - The error message
  * @property {Object} context - the request, user and report context
  * @property {Object} context.httpRequest - the request context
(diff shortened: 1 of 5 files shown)


cristiancavalli:
@matthewloring @ofrobots PTAL

lib/classes/error-message.js
```diff
@@ -67,7 +67,7 @@ var isObject = lodash.isObject;
 function ErrorMessage() {
 
   this.eventTime = (new Date()).toISOString();
-  this.serviceContext = {service : '', version : ''};
+  this.serviceContext = {service : 'node', version : undefined};
```
ofrobots: Is it preferable to define `version` to the value `undefined` or would it be better to omit the key `version` from serviceContext altogether? What does the errors API prefer?

ofrobots:
LGTM w/ a question.

steren:
I believe it is better to just not set the version.

cristiancavalli:
@ofrobots @steren Since it is jsonified the version key, if left undefined, will be omitted from from the JSON string:

screen shot 2016-08-24 at 2 00 18 pm

This allows the key to be present in the data sink for clarity of the total domain yet transparent if left unset by the user.

PR #55, opened 2016-08-25
Add manual end-to-end test fixture

cristiancavalli:
Add a e2e test using the manual interface. Add delete/get API
interfaces as extensions of the auth client for presence checking
of transmitted errors. Set e2e fixture to run after integration
and unit suite.

Notes:
The e2e fixture is running extraordinarily slowly but that is
because of the 120000ms timeouts which must take place between
remote deletes/creates and reads.

### DIFF
--- a/tests/e2e/manual-interface.js
+++ b/tests/e2e/manual-interface.js
@@ -1,0 +1,75 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+var lodash = require('lodash');
+var isObject = lodash.isObject;
+var isEmpty = lodash.isEmpty;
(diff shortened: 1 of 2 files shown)


cristiancavalli:
@GoogleCloudPlatform/node-team PTAL

tests/e2e/manual-interface.js
```diff
@@ -0,0 +1,75 @@
...
+              if (err) {
+                t.fail('Could not get the groups for test');
+                console.log(err);
+                t.end();
+                return;
+              }
+              t.assert(!isEmpty(groups), 'Groups should not be empty');
+              t.end();
+            });
+          }, 1200);
```
ofrobots: Move the constant to a var at the top of the file for easy timeout tweaking in the future.

ofrobots: Is 1.2seconds long enough for the errors service to be able to reflect updates back to us?

ofrobots: ping

cristiancavalli: @ofrobots I can't find the docs indicating the SLA, 1.2 passes my integration tests consistently

ofrobots: I think you should use at least 20 seconds as the delay to make sure we don't have flaky test failures.

Also, I think this test might need to be updated to be a mocha test now.

ofrobots: Example of a system-test from google-cloud-logging: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/logging-winston/system-test/logging-winston.js#L30

DominicKramer:
This PR has been closed since its contents have been incorporated into the [google-cloud-node/error-reporting](https://github.com/GoogleCloudPlatform/google-cloud-node/tree/master/packages/error-reporting) repository via [PR #2264](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2264).

Issue #56, opened 2016-08-25
Detect env on uncaught unit test for fork, update test scripts


cristiancavalli:
It's failing right now, but not in the way that it's supposed to on non-internal merges.

PR #57, opened 2016-08-25
Update testUncaught.js with env detection

cristiancavalli:
Check for the `GOOGLE_APPLICATION_CREDENTIALS` and `GCLOUD_PROJECT` environmental variables before running the fixture in the unit suite. If the variables are not set skip the test.

Related issues:
Fixes: #56

### DIFF
--- a/tests/unit/testUncaught.js
+++ b/tests/unit/testUncaught.js
@@ -15,7 +15,9 @@
  */
 
 var test = require('tape');
-var isFunction = require('lodash').isFunction;
+var lodash = require('lodash');
+var isFunction = lodash.isFunction;
+var isString = lodash.isString;
 var uncaughtSetup = require('../../lib/interfaces/uncaught.js');
 var Configuration = require('../../lib/configuration.js');
 var ErrorMessage = require('../../lib/classes/error-message.js');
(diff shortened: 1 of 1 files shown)


tests/unit/testUncaught.js
```diff
@@ -50,6 +52,13 @@ test('Uncaught handler setup', function (t) {
 });
 
 test('Test uncaught shutdown behavior', function (t) {
+  if (!isString(process.env.GOOGLE_APPLICATION_CREDENTIALS)
```
matthewloring: It may be better to fail if the variables aren't set. If this is being run automatically on ci, this test could start skipping without being detected unless the output is inspected manually (unless maybe travis reports skipped tests in some way)?

cristiancavalli: @matthewloring Wouldn't this then cause the ci to always fail on pr's arising externally? Istanbul does log skipped tests as pending, not sure if that is enough in this case though.

matthewloring: Ah, that is a good point. Why is `process.env.GOOGLE_APPLICATION_CREDENTIALS` required in this test again? If we aren't actually publishing maybe it and `GCLOUD_PROJECT` could be set to dummy values for testing only as is done in the trace agent?

cristiancavalli: @matthewloring Good point - i'll look into stubbing these out

cristiancavalli: @matthewloring on second look - the reason this is required is becuase the second part of this test utilizes the google-auth-library which needs `GOOGLE_APPLICATION_CREDENTIALS` set (or some other valid form of configuration) so that it won't throw

matthewloring: Ok, I'm fine leaving this in for now. It would be good to avoid if possible in the future as we have had skipped tests start failing unnoticed in the past.

matthewloring:
LGTM.

PR #63, opened 2016-10-12
Add error stack substitution to string handler

### LINKED ISSUE #41: report(string) does not report errors
calling report with a simple string does not report errors in Error Reporting.
Example code I used on GAE Flex:
```JS
app.get('/', function (req, res) {
  errorHandler.report('simple text test steren');
  res.status(200).send('ok');
});
```

The reason might be because of what is sent to the API:
If you are not sending a stacktrace in the message attribute, then you must provide context.reportLocation otherwise the backends are not able to group the errors and will ignore it.
See https://cloud.google.com/error-reporting/docs/formatting-error-messages

for the client-side JavaScript library, I actually always create an Error from the message string: see https://github.com/GoogleCloudPlatform/stackdriver-errors-js/blob/master/stackdriver-errors.js#L87

cristiancavalli:
Add logic to provide error stacks for basic non-
error string tokens given as error values. Update
the test-servers/manual_scaffold_server.js test
file to use this new pattern.

Related Issues:
Fixes #41

### DIFF
--- a/lib/error-handlers/string.js
+++ b/lib/error-handlers/string.js
@@ -30,12 +30,16 @@ var isString = require('lodash').isString;
  */
 function handleStringAsError(err, errorMessage) {
   var fauxError = new Error();
-  var errChecked = fauxError.stack;
+  var fullStack = fauxError.stack.split('\n');
+  var cleanedStack = fullStack.slice(0, 1).concat(fullStack.slice(4));
+  var errChecked = '';
 
   if (isString(err)) {
-
-    errChecked = err;
+    // Replace the generic error message with the user-provided string
+    cleanedStack[0] = err;
   }
+
+  errChecked = cleanedStack.join('\n');
 
   errorMessage.setMessage(errChecked);
 }
--- a/tests/test-servers/manual_scaffold_server.js
+++ b/tests/test-servers/manual_scaffold_server.js
@@ -14,8 +14,12 @@
  * limitations under the License.
  */
 var errors = require('../../index.js').start();
-var r = errors.report('Sample string', (err, result) => {
-  console.log('callback from report', err, result);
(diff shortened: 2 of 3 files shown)


steren:
This solution LGTM.

cristiancavalli:
Changes:
- Added default of `nodejs` to `configuration->serviceContext->service`.
- Amended to current commit and update commit message.

tests/test-servers/manual_scaffold_server.js
```diff
@@ -14,8 +14,12 @@
  * limitations under the License.
  */
 var errors = require('../../index.js').start();
-var r = errors.report('Sample string', (err, result) => {
-  console.log('callback from report', err, result);
+var r = errors.report('Sample test string', (err, response, body) => {
```
matthewloring: Could this be made into a test that will fail if this no longer works?

lib/error-handlers/string.js
```diff
@@ -30,13 +30,17 @@ var isString = require('lodash').isString;
  */
 function handleStringAsError(err, errorMessage) {
   var fauxError = new Error();
-  var errChecked = fauxError.stack;
+  var fullStack = fauxError.stack.split('\n');
+  var cleanedStack = fullStack.slice(0, 1).concat(fullStack.slice(4));
```
matthewloring: Slicing out particular numbers of stack frames could become incorrect as new functions are added. Could we add a test checking that the correct number of stack frames are omitted?

tests/test-servers/manual_scaffold_server.js
cristiancavalli: Will provide as follow-on

lib/error-handlers/string.js
cristiancavalli: Will provide as follow-on

PR #64, opened 2016-11-01
Test Fixes

kjin:
The Travis CI build is incorrectly configured to swallow the `tape` exit code, so it always passes.

This change addresses this by modifying the Travis config file and fixing broken tests that went undetected.

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -87,7 +87,7 @@ var Configuration = function(givenConfig, logger) {
    * @type {Boolean}
    */
   this._shouldReportErrorsToAPI = env.NODE_ENV === 'production';
-  if (!this._shouldReportErrorsToAPI) {
+  if (!this._shouldReportErrorsToAPI && this._logger) {
     this._logger.warn([
       'Stackdriver error reporting client has not been configured to send',
       'errors, please check the NODE_ENV environment variable and make sure it',
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -20,7 +20,6 @@ var commonDiag = require('@google/cloud-diagnostics-common');
 var utils = commonDiag.utils;
 var lodash = require('lodash');
 var isPlainObject = lodash.isPlainObject;
-var isFunction = lodash.isFunction;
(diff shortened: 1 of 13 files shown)


kjin:
Request for feedback pending on Travis CI build result and output.

lib/configuration.js
```diff
@@ -87,7 +87,7 @@ var Configuration = function(givenConfig, logger) {
    * @type {Boolean}
    */
   this._shouldReportErrorsToAPI = env.NODE_ENV === 'production';
-  if (!this._shouldReportErrorsToAPI) {
```
matthewloring: In what cases would we expect `_logger` to not be defined at this point? If this is for testing only, I'd rather use a dummy logger for testing instead of adding logic to skip this warning. @cristiancavalli

matthewloring:
Thanks for catching this!

lib/configuration.js
kjin: Makes sense, since Configuration is never instantiated by the user.

kjin:
PTAL

lib/configuration.js
matthewloring: In that case, could we put together a dummy logger for the tests and remove the check?

kjin: See 6575dd6

tests/fixtures/configuration.js
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var Configuration = require('../../lib/configuration.js');
+
+var FakeConfiguration = function(config) {
+  return Configuration.bind(this)(config, { warn: function () {} });
```
matthewloring: Prefer: `Configuration.call(this, config, { warn: function() {} })` to immediately invoking a bound function.

cristiancavalli:
LGTM

PR #68, opened 2016-11-09
Omit Koa when ES6 Generators are not supported

kjin:
This change prevents the Koa error middleware from loading in environments where ES6 generators are not supported (namely, Node 0.12 without the `--harmony` flag).

Also see: https://github.com/GoogleCloudPlatform/cloud-errors-nodejs/issues/65


kjin:
PTAL

PR #70, opened 2016-11-11
The service defaults to the function name on GCF

DominicKramer:
In particular, if the user has specified a service name, then
that value will be used as the service name.  Otherwise, if the
code is running on GCF, the name of the function being run will
be used as ther service name.  Last, if neither of those two
conditions are true, the GCP service name will be used.


DominicKramer:
PTAL.  Before continuing, I want to make sure the resolution order of the service name is correct.

lib/configuration.js
```diff
@@ -174,7 +175,19 @@ var Configuration = function(givenConfig, logger) {
  * @returns {Undefined} - does not return anything
  */
 Configuration.prototype._checkLocalServiceContext = function() {
-  var gaeModuleName = env.GAE_MODULE_NAME;
+  // Note: The GAE_MODULE_NAME environment variable is set on GCP.
```
ofrobots: GAE_MODULE_NAME is only set on GAE (AppEngine).

lib/configuration.js
```diff
@@ -174,7 +175,19 @@ var Configuration = function(givenConfig, logger) {
...
+  // To determine the service name to use:
+  //   If the user specified a service name it should be used, otherwise 
+  //   if the FUNCTION_NAME environment variable is set (indicating that the 
+  //   code is running on GCF) then the FUNCTION_NAME value should be used as 
+  //   the service name.  If neither of these conditions are true, the 
+  //   value of the GAE_MODULE_NAME environment variable should be used as the 
+  //   service name.
+  var funcName = env.FUNCTION_NAME;
+  var gaeModuleName = isUndefined(funcName) ? env.GAE_MODULE_NAME : funcName;
   var gaeModuleVer = env.GAE_MODULE_VERSION;
```
ofrobots: Does GCF have a concept of function version? Intuitively `GAE_MODULE_VERSION` should be used iff we also are going to use `GAE_MODULE_NAME` as the service name.

lib/configuration.js
```diff
@@ -174,7 +175,19 @@ var Configuration = function(givenConfig, logger) {
...
+  //
+  // To determine the service name to use:
+  //   If the user specified a service name it should be used, otherwise 
+  //   if the FUNCTION_NAME environment variable is set (indicating that the 
+  //   code is running on GCF) then the FUNCTION_NAME value should be used as 
+  //   the service name.  If neither of these conditions are true, the 
+  //   value of the GAE_MODULE_NAME environment variable should be used as the 
+  //   service name.
+  var funcName = env.FUNCTION_NAME;
+  var gaeModuleName = isUndefined(funcName) ? env.GAE_MODULE_NAME : funcName;
```
ofrobots: The variable name doesn't reflect what it contains. A better variable name would be `service`.

tests/unit/testServiceConfiguration.js
```diff
@@ -0,0 +1,160 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+var test = require('tape');
+var lodash = require('lodash');
+var isNumber = lodash.isNumber;
+var Configuration = require('../fixtures/configuration.js');
+var level = process.env.GCLOUD_DEBUG_LOGLEVEL;
```
ofrobots: Wrong environment variable?

tests/unit/testServiceConfiguration.js
```diff
@@ -0,0 +1,160 @@
...
+    process.env.FUNCTION_NAME = oldFunctionName;
+    process.env.GAE_MODULE_NAME = oldModuleName;
+
+    t.end();
+  }
+);
+
+test(
+  'A Configuration uses the service name in the given config if it ' +
+    'was specified and only the GEA_MODULE_NAME env var is set',
```
kjin: Nit: GAE_MODULE_NAME instead of GEA (this occurs multiple times)

DominicKramer: A typo on my part.  Thanks for catching it.

lib/configuration.js
DominicKramer: Fixed

tests/unit/testServiceConfiguration.js
DominicKramer: Fixed to use `GCLOUD_ERRORS_LOGLEVEL` instead of `GCLOUD_DEBUG_LOGLEVEL`.  I also fixed this typo in the `tests/unit/testConfiguration.js` file.

lib/configuration.js
DominicKramer: Updated the variable name to be `service`.

lib/configuration.js
DominicKramer: Fixed.  The unit tests verify the different combinations of having `GAE_MODULE_NAME`, `GAE_MODULE_VERSION`, `FUNCTION_NAME` and/or user specified context name and version information specified.

steren:
Can you confirm that in the case when the user does not provide a service name and neither GAE_MODULE_NAME nor FUNCTION_NAME are set, then we use a default service called "node"?
I think this was the behavior discussed in the past. I'd suggest to add a test about this.

DominicKramer:
@steren @ofrobots I have updated the code to use 'node' as the default service name if neither the `FUNCTION_NAME` nor `GAE_MODULE_VERSION` environment variables are set and the user has not specified a service name.  I have also included unit tests for this case that verify the service name as well as the service version.

With that in mind, what should the fallback service version be?  In the code as it is, the empty string is used as the default service version.

In particular, if the user specifies a service version, it is used.  Otherwise if the user doesn't specify a service version and either the `FUNCTION_NAME` environment variable is set or neither the `FUNCTION_NAME` nor the `GAE_MODULE_NAME` environment variables are set, the default service version is used.

Does this sound like the correct behavior?  Also, is an empty string a good default service version?

steren:
If no version is available in the environment or provided by the user, you should just not set serviceContext.version (not even an empty string).
Our system behaves correctly when no version is set. The errors will be aggregated at the service level, and the UI will disable the version picker for this service.

Here is the order of priority to recap:

1. user defined serviceContext.version
2. `GAE_MODULE_VERSION` if present
3. nothing

In your previous message, there seem to be confusion between `GAE_MODULE_VERSION` and `GAE_MODULE_NAME`. Let me look at the tests and add comments.

lib/configuration.js
```diff
@@ -208,11 +208,25 @@ Configuration.prototype._checkLocalServiceContext = function() {
...
+  if (isUndefined(funcName)) {
+    service = env.GAE_MODULE_NAME;
+    version = env.GAE_MODULE_VERSION;
+  } else {
+    service = funcName;
+  }
+
   this._serviceContext.service = isString(service) ? service : '';
-  this._serviceContext.version = isString(gaeModuleVer) ? gaeModuleVer : '';
+  this._serviceContext.version = isString(version) ? version : '';
```
steren: Leave not set if no version available, no need for empty string.

tests/unit/testServiceConfiguration.js
```diff
@@ -0,0 +1,397 @@
...
+    process.env.GAE_MODULE_NAME = oldModuleName;
+
+    t.end();
+  }
+);
+
+test(
+  'A Configuration uses the service name "node" and a blank version if ' +
+  'GAE_MODULE_NAME is not set, FUNCTION_NAME is not set, and the user has '+
+  'not specified a service name or version even if GAE_MODULE_VERSION has ' +
```
steren: This case should not happen in theory. 
I am OK with the describe behavior. I would also be OK if the `GAE_MODULE_VERSION` was recorded as serviceContect.version.

cristiancavalli:
LGTM

lib/configuration.js
```diff
@@ -196,10 +196,37 @@ var Configuration = function(givenConfig, logger) {
...
+  //   Otherwise, the value of the environment variable GAE_MODULE_VERSION
+  //   will be used if and only if the FUNCTION_NAME environment variable is
+  //   not set.
+  var service;
+  var version;
+
+  if (env.FUNCTION_NAME){
+    service = env.FUNCTION_NAME;
+  }
+  else if (env.GAE_MODULE_NAME){
```
ofrobots: AppEngine Flex has starting using a new set of environment variables. See: https://cloud.google.com/appengine/docs/flexible/python/migrating. It would be good to start accepting them as well here.

lib/configuration.js
DominicKramer: I've updated the version to default to being undefined instead of the empty string.

tests/unit/testServiceConfiguration.js
DominicKramer: I've updated this to have an undefined version instead of the empty string.

lib/configuration.js
DominicKramer: @ofrobots I've updated the code to also accept `GAE_SERVICE` and `GAE_VERSION`.  These two variables have precedence over `GAE_MODULE_NAME` and `GAE_MODULE_VERSION` but not the user configuration.

tests/unit/testMigratedServiceConfiguration.js
```diff
@@ -0,0 +1,553 @@
...
+    'if the service name is not given in the given config ' +
+    'even if the GAE_SERVICE was not set',
+  function(t) {
+    var oldServiceName = process.env.GAE_SERVICE || '';
+    var oldServiceVersion = process.env.GAE_VERSION || '';
+
+    var oldModuleName = process.env.GAE_MODULE_NAME || '';
+    var oldModuleVersion = process.env.GAE_MODULE_VERSION || '';
+
+    var oldFunctionName = process.env.FUNCTION_NAME || '';
```
ofrobots: Extract this boiler plate into a standalone `before` function to avoid repetition, and likewise for `after`.

PR #72, opened 2016-11-15
Add error log when credentials cannot be identified

### LINKED ISSUE #2: User should be warned when key has not been provided
I just tried the agent in a sample app running locally where I didn't provide a key in the config or in through the environment. The error reporting agent cannot publish errors in this case (expected) but it doesn't provide any indication to the user that it has been unsuccessful at publishing and the user is left wondering whether the errors are actually going to show up on the Stackdriver console.

```
❯ node index.js
Example app listening!
SyntaxError: Unexpected token :
    at Object.parse (native)
    at /Users/ofrobots/src/node-hello/index.js:15:8
    at Layer.handle [as handle_request] (/Users/ofrobots/src/node-hello/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/ofrobots/src/node-hello/node_modules/express/lib/router/route.js:131:13)
    at Route.dispatch (/Users/ofrobots/src/node-hello/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/Users/ofrobots/src/node-hello/node_modules/express/lib/router/layer.js:95:5)
    at /Users/ofrobots/src/node-hello/node_modules/express/lib/router/index.js:277:22
    at Function.process_params (/Users/ofrobots/src/node-hello/node_modules/express/lib/router/index.js:330:12)
    at next (/Users/ofrobots/src/node-hello/node_modules/express/lib/router/index.js:271:10)
    at expressInit (/Users/ofrobots/src/node-hello/node_modules/express/lib/middleware/init.js:33:5)
^C
```

I hit Ctrl-C after a minute. I think it would be useful if we could tell user about the lack of authentication information. The sooner the better, i.e. it would be best if we didn't have to wait till the first error before we told the user that their errors won't be reported to the API.

cristiancavalli:
Add a private member function to the Configuration class
to be called after gathering any available local configuration
values. The function checks to see if at least one of the three credential
fields has been set. If this is not the case then the function will write
an error log telling the user that credentials have not been
supplied and authorization with the Stackdriver API will not be attainable
until such time as they are provided.

Fixes #2

### DIFF
--- a/src/configuration.js
+++ b/src/configuration.js
@@ -162,6 +162,15 @@ var Configuration = function(givenConfig, logger) {
    */
   this._version = version;
   /**
+    * Boolean flag indicating whether or not the configuration instance was able
+    * to find a set of usable credentials to attempt authorization against the
+    * Stackdriver API.
+    * @memberof Configuration
+    * @private
+    * @type {Boolean}
+    */
+  this._lacksValidCredentials = true;
+  /**
    * The _givenConfiguration property holds a ConfigurationOptions object
    * which, if valid, will be merged against by the values taken from the meta-
    * data service. If the _givenConfiguration property is not valid then only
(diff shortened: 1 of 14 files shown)


cristiancavalli:
@ofrobots @matthewloring PTAL when you get a sec

lib/configuration.js
```diff
@@ -250,6 +250,27 @@ Configuration.prototype._gatherLocalConfiguration = function() {
...
+ * @returns {Undefined} - does not return anything
+ */
+Configuration.prototype._checkAuthConfiguration = function () {
+  if (isNull(this._key) && isNull(this._keyFilename) && 
+    isNull(this._credentials) && 
+    isEmpty(process.env.GOOGLE_APPLICATION_CREDENTIALS)) {
+      this._logger.error([
+        'Unable to find credential information on instance. This library will',
+        'be unable to communicate with the Stackdriver API to save errors.'
+      ].join(' '));
```
matthewloring: Do we need to take any additional action here to disable the agent in the case where it cannot make authenticated requests?

cristiancavalli: @matthewloring Handlers have been updated to defer error parsing and send logic if configuration lacks usable credentials.

lib/configuration.js
```diff
@@ -167,6 +167,15 @@ var Configuration = function(givenConfig, logger) {
...
   this._version = version;
   /**
+   * Boolean flag indicating whether or not the configuration instance was able
+   * to find a set of usable credentials to attempt authorization against the
+   * Stackdriver API.
+   * @memberof Configuration
+   * @private
+   * @type {Boolean}
+   */
+  this._lacksValidCredentials = false;
```
matthewloring: Is it possible for this value to be observed before a call to `_checkAuthConfiguration`? Since this defaults to `false` errors will think valid credentials are available if they have not yet been checked. Maybe it would be better to default to `true`.

cristiancavalli: Not sure how that would work -- it's _checkAuthConfiguration that is responsible for setting the value of _lacksValidCredentials to true if it cannot find a valid set of credentials. The pr has been updated to default to true though

PR #74, opened 2016-11-22
Add error message builder and parse capability

cristiancavalli:
Add a function exposed on the top-level export
of the library which allows for creation of new
ErrorMessage class instances for direct use and
modification by application code.

Additionally, add capability to the manual handler
to specifically accept error message class instances
and not apply general object handling. This means
that if a user supplies only an error message instance
the instance will be accepted wholesale and queued for
submission to the service without modification.

Example invocation:

```JS
var errors = require(...).start();

var em = errors.event()
  .setMessage('three token string')
  .setUser('user');
// The same variadic argument structure applies
errors.report(em, () => console.log('done!'));
```

@ofrobots @matthewloring  ptal

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,6 +32,7 @@ var hapi = require('./src/interfaces/hapi.js');
 var manual = require('./src/interfaces/manual.js');
 var express = require('./src/interfaces/express.js');
 var restify = require('./src/interfaces/restify');
+var messageBuilder = require('./src/interfaces/message-builder.js');
 var uncaughtException = require('./src/interfaces/uncaught.js');
 var createLogger = require('./src/logger.js');
 
@@ -65,6 +66,7 @@ var createLogger = require('./src/logger.js');
  *  Stackdriver Error Reporting Service
  * @property {Function} express - The express plugin for Stackdriver Error
  *  Reporting
+ * @property {Function} message - Returns a new ErrorMessage class instance
  */
 
 // TODO: Update this documentation
@@ -96,6 +98,7 @@ function Errors(initConfiguration) {
   this.report = manual(client, config);
   this.express = express(client, config);
   this.restify = restify(client, config);
+  this.event = messageBuilder(config);
 
   if (koa) {
     this.koa = koa(client, config);
--- a/README.md
+++ b/README.md
@@ -125,6 +125,21 @@ var errors = require('@google/cloud-errors')({
 
 ## Examples
 
+### Reporting Manually
+
+```JS
+var errors = require('@google/cloud-errors').start();
+// Use the error message builder to custom set all message fields
+var errorEvt = errors.event()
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -31,9 +31,11 @@ var koa = useKoa ? require('./lib/interfaces/koa.js') : null;
 var hapi = require('./lib/interfaces/hapi.js');
 var manual = require('./lib/interfaces/manual.js');
 var express = require('./lib/interfaces/express.js');
-var restify = require('./lib/interfaces/restify');
+var restify = require('./lib/interfaces/restify.js');
+var messageBuilder = require('./lib/interfaces/message-builder.js');
 var uncaughtException = require('./lib/interfaces/uncaught.js');
 var createLogger = require('./lib/logger.js');
+var ErrorMessage = require('./lib/classes/error-message.js');
```
matthewloring: Is this used?

lib/interfaces/message-builder.js
```diff
@@ -0,0 +1,44 @@
...
+ *  configuration
+ * @returns {ErrorMessage} - a new ErrorMessage instance
+ */
+function handlerSetup(config) {
+  /**
+   * The interface for creating new instances of the ErrorMessage class which
+   * can be used to send custom payloads to the Error reporting service.
+   * @returns {ErrorMessage} - returns a new instance of the ErrorMessage class
+   */
+  function newMessage() {
```
matthewloring: It may be cleaner to return this function directly since it is only referenced once.

cristiancavalli: @matthewloring It's using a closure so that it can retain access to config with out becoming a class.

index.js
cristiancavalli: Thanks! updated

lib/interfaces/message-builder.js
matthewloring: I was just thinking it would be simpler to write:

```javascript
return function() { ... };
```

instead of

```javascript
function foo { ... }
return foo;
```

Just a matter of preference.

steren:
I find the class name `ErrorMessage` confusing, especially when you do:
````
errors.message()
  .setMessage('three token string')
  .setUser('user');
```
Could you use `ErrorEvent`? This will be more aligned with the terms we use in the API and documentation.

cristiancavalli:
@steren Signature updated - wdyt?

steren:
LGTM

cristiancavalli:
@ofrobots updated

ofrobots:
LGTM once the tests are green.

cristiancavalli:
@ofrobots CIs passing, going to let pend and do a second once-over tonight before merging into master

ofrobots:
Still LGTM.

PR #78, opened 2016-11-30
Specify lodash dependencies over full library

cristiancavalli:
Specify only required lodash submodules in both
dependencies and development dependencies.


cristiancavalli:
@ofrobots @matthewloring ptal

ofrobots:
`request` depends on full `lodash` so wouldn't actually end up increasing the disk footprint?

cristiancavalli:
@ofrobots I don't see that in the deps on the GH page:
https://github.com/request/request/blob/master/package.json

Am I missing it?

ofrobots:
Apologies. The dependency chain as per `yarn`: `@google/cloud-errors#@google/cloud-diagnostics-common#google-auth-library#request#form-data#async`

cristiancavalli:
Ahh, my bad. I'll close this pr.

ofrobots:
Thanks for putting this together though!

PR #80, opened 2016-12-01
Add force flag option to configuration

### LINKED ISSUE #79: Provide an option to override NODE_ENV
I see the `NODE_ENV === 'production'` condition is hardcoded at the moment, which seems pretty limiting (e.g. can't use error reporting on staging). And, obviously, changing NODE_ENV to toggle error reporting might not be a suitable option for some cases.

cristiancavalli:
Add a optional boolean flag option to the configuration which
allows for the configuration of the library to send errors to
the Stackdriver service regardless of env configuration.

fixes #79

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -77,22 +77,16 @@ var Configuration = function(givenConfig, logger) {
    * the Stackdriver error reporting library will actually try to report Errors
    * to the Stackdriver Error API. The value of this property is derived from
    * the `NODE_ENV` environmental variable. If the `NODE_ENV` variable is set to
-   * 'production' then the _shouldReportErrorToAPI property will be set to true
+   * 'production' then the _shouldReportErrorsToAPI property will be set to true
    * and error reporting library will attempt to send errors to the Error API.
    * Otherwise the value will remain false and errors will not be reported to
    * the API.
    * @memberof Configuration
    * @private
    * @type {Boolean}
+   * @defaultvalue false
    */
-  this._shouldReportErrorsToAPI = env.NODE_ENV === 'production';
-  if (!this._shouldReportErrorsToAPI) {
-    this._logger.warn([
-      'Stackdriver error reporting client has not been configured to send',
-      'errors, please check the NODE_ENV environment variable and make sure it',
-      'is set to production'
-    ].join(' '));
-  }
+  this._shouldReportErrorsToAPI = false;
   /**
    * The _projectId property is meant to contain the string project id that the
    * hosting application is running under. The project id is a unique string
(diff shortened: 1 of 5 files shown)


cristiancavalli:
@steren @ofrobots PTAL

lib/configuration.js
```diff
@@ -229,6 +223,20 @@ Configuration.prototype._checkLocalServiceContext = function() {
...
+  } else if (has(this._givenConfiguration, 'force')) {
+    throw new Error('config.force must be a boolean');
+  } else {
+    this._shouldReportErrorsToAPI = env.NODE_ENV === 'production';
+  }
+  if (!this._shouldReportErrorsToAPI) {
+    this._logger.warn([
+      'Stackdriver error reporting client has not been configured to send',
+      'errors, please check the NODE_ENV environment variable and make sure it',
+      'is set to production'
```
steren: also mention the `force` attribute here (like you do in the other error message)

lib/google-apis/auth-client.js
```diff
@@ -125,7 +125,8 @@ RequestHandler.prototype.sendError = function(errorMessage, userCb) {
     cb(new Error([
       'Stackdriver error reporting client has not been configured to send',
       'errors, please check the NODE_ENV environment variable and make sure it',
-      'is set to production'
+      'is set to production or set the force property to true in the runtime',
```
steren: maybe decorate production and other attributes with ":
is set to "production" or set the "force" property to true in the runtime

lib/configuration.js
cristiancavalli: @steren thanks for the catch! updated

lib/google-apis/auth-client.js
cristiancavalli: @steren good point - decorated with double quotes

lib/configuration.js
```diff
@@ -229,6 +223,21 @@ Configuration.prototype._checkLocalServiceContext = function() {
...
+  } else if (has(this._givenConfiguration, 'force')) {
+    throw new Error('config.force must be a boolean');
+  } else {
+    this._shouldReportErrorsToAPI = env.NODE_ENV === 'production';
+  }
+  if (!this._shouldReportErrorsToAPI) {
+    this._logger.warn([
+      'Stackdriver error reporting client has not been configured to send',
+      'errors, please check the NODE_ENV environment variable and make sure it',
+      'is set to production or the force property is set to true in the',
```
steren: add double quotes here too.

cristiancavalli: @steren lol thanks for catch again -- still getting my first coffee on over here

ofrobots:
Bike-shed time! /cc @matthewloring @kjin @DominicKramer.

IMO `force` is not descriptive enough – what is it forcing?. Alternative suggestions: forceEnable, ignoreEnvironmentCheck, ...

matthewloring:
`ignoreEnvironmentCheck` sound good. `ignoreNodeEnv` may be more specific. `reportInDevelopment` also came to mind.

cristiancavalli:
Updated with config property name as `ignoreEnvironmentCheck`

lib/configuration.js
```diff
@@ -229,6 +223,21 @@ Configuration.prototype._checkLocalServiceContext = function() {
...
+    throw new Error('config.ignoreEnvironmentCheck must be a boolean');
+  } else {
+    this._shouldReportErrorsToAPI = env.NODE_ENV === 'production';
+  }
+  if (!this._shouldReportErrorsToAPI) {
+    this._logger.warn([
+      'Stackdriver error reporting client has not been configured to send',
+      'errors, please check the NODE_ENV environment variable and make sure it',
+      'is set to "production" or the ignoreEnvironmentCheck property is set to',
+      '"true" in the runtime configuration object'
```
ofrobots: nit: drop the double quotes around true as people might set it to the string "true".

lib/configuration.js
```diff
@@ -229,6 +223,21 @@ Configuration.prototype._checkLocalServiceContext = function() {
  * @returns {Undefined} - does not return anything
  */
 Configuration.prototype._gatherLocalConfiguration = function() {
+  if (this._givenConfiguration.ignoreEnvironmentCheck === true) {
+    this._shouldReportErrorsToAPI = true;
+  } else if (has(this._givenConfiguration, 'ignoreEnvironmentCheck')) {
```
ofrobots: nit: It would be better to do an `isBoolean` check as the first `if`.

lib/google-apis/auth-client.js
```diff
@@ -125,7 +125,8 @@ RequestHandler.prototype.sendError = function(errorMessage, userCb) {
     cb(new Error([
       'Stackdriver error reporting client has not been configured to send',
       'errors, please check the NODE_ENV environment variable and make sure it',
-      'is set to production'
+      'is set to "production" or set the ignoreEnvironmentCheck property to ',
+      '"true" in the runtime configuration object'
```
ofrobots: ditto

PR #81, opened 2016-12-04
Update configuration.js, README.md

cristiancavalli:
Update description of `_shouldReportErrorsToAPI` and how value is derived
in `configuration.js`. Update line in `README.md` stating that only a production
`NODE_ENV` variable was required for service error reporting to also include 
that new runtime property `ignoreEnvironmentCheck` can be used to override
environment-wide runtime configurations..
Update sample runtime configuration snippet to be more consistent
in describing some properties and their default values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ applications running in almost any environment. Here's an introductory video:
 1. Your application needs to use Node.js version 0.12 or greater.
 1. You need a [Google Cloud project](https://console.cloud.google.com). Your application can run anywhere, but errors are reported to a particular project.
 1. [Enable the Stackdriver Error Reporting API](https://console.cloud.google.com/apis/api/clouderrorreporting.googleapis.com/overview) for your project.
-1. The module will only send errors when the `NODE_ENV` environment variable is set to `production`.
+1. The module will only send errors when the `NODE_ENV` environment variable is 
+set to `production` or the `ignoreEnvironmentCheck` property given in the
+runtime configuration object is set to `true`.
 
 ## Quick Start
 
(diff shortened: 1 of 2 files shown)


cristiancavalli:
@GoogleCloudPlatform/node-team ptal

cristiancavalli:
@matthewloring sorry, found another overlooked reference about NODE_ENV in the readme. Now updated.

README.md
```diff
@@ -22,7 +22,6 @@ applications running in almost any environment. Here's an introductory video:
 1. Your application needs to use Node.js version 0.12 or greater.
 1. You need a [Google Cloud project](https://console.cloud.google.com). Your application can run anywhere, but errors are reported to a particular project.
 1. [Enable the Stackdriver Error Reporting API](https://console.cloud.google.com/apis/api/clouderrorreporting.googleapis.com/overview) for your project.
-1. The module will only send errors when the `NODE_ENV` environment variable is set to `production`.
```
matthewloring: I was going to suggest that this be modified to mention that we either need `NODE_ENV` or the `ignoreEnvironmentCheck` configuration to send errors but it looks like `ignoreEnvironmentCheck` is true by default. My understanding was that we still wanted to only report errors in production by default but allow a mechanism to bypass this behavior (looks like I missed this in the last code review). Do we want to default to reporting all errors now?

cristiancavalli: ignoreEnvironmentCheck is not true by default -- I'll update the snippet and expand the line

cristiancavalli: @matthewloring updated

matthewloring: Ah, sorry for the confusion.

cristiancavalli:
@matthewloring no worries, snippet was confusing and needed some cleaning up

PR #82, opened 2016-12-05
Update integration tests

cristiancavalli:
Prefer use of newly available `ignoreEnvironmentCheck`
configuration property instead of relying on `NODE_ENV`
checks to better control success and failure conditions
of integration tests.


cristiancavalli:
@matthewloring @ofrobots ptal

Issue #10, opened 2016-07-12
Errors not sent when providing an API key only


steren:
- I ran `export NODE_ENV=production`
- I installed express and cloud-errors.
- I run the following code:

I could not see errors in the Error Reporting UI for this project.
I could successfully send manually errors to this project with this API key (by calling the API from cURL).

```
var express = require('express');
var app = express();
var errorHandler = require('@google/cloud-errors')({
  projectId: "<my project>",
  key: "<my api key>",
  serviceContext: {
      service: 'steren-test-node',
      version: 'test1'
  }
});

app.get(
  '/errorRoute',
  function ( req, res, next ) {
    // You can push in errors manually
    res.send("Error");
    res.end();
    next(new Error("Got traffic on the errorRoute"));
  }
);

app.get(
  '/anotherRoute',
  function ( req, res, next ) {
    // It'll even log potentially unexpected errors
    JSON.parse("{\"malformedJson\": true");
  }
)

// Just use the express plugin
app.use(errorHandler.express);

app.listen(
  3000
  , function ( ) {
    console.log('Server has been started on port 3000');
  }
);
```

steren:
I now tested the module with Vanilla Node.js:

I ran `export NODE_ENV=production`

My code:
```
var errorHandler = require('@google/cloud-errors')({
    projectId: '<project id>',
    key: '<valid API key>'
});

errorHandler.report(
    new Error('This is a test'),
    (err, res) => {

        if (err) {
            console.log('Error was unable to be reported', err);
        } else {
            console.log('Error reported!', res);
        }
    }
);
```

I get this error:

>  $ node index.js
> Error was unable to be reported { [Error: Stackdriver Error Reporting API has not been used in project usable-auth-library before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/clouderrorreporting.googleapis.com/overview?project=usable-auth-library then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.]
>   code: 403,
>   errors:
>    [ { message: 'Stackdriver Error Reporting API has not been used in project usable-auth-library before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/clouderrorreporting.googleapis.com/overview?project=usable-auth-library then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.',
>        domain: 'global',
>        reason: 'forbidden' } ] }

[...]

PR #84, opened 2016-12-06
Remove API Key instructions from README

steren:
While #10 is opened, we should not document the support of API keys.

@cristiancavalli PTAL


cristiancavalli:
LGTM

PR #85, opened 2016-12-09
Rename the `lib` directory to `src`

DominicKramer:


DominicKramer:
PTAL @cristiancavalli Do you think the `config.tar.enc` file needs to be regenerated?

cristiancavalli:
LGTM

PR #87, opened 2016-12-12
The module now exports a constructor

DominicKramer:
Now this module exports a constructor.  This change was made to more closely align this module with the other Google Cloud API Client Libraries via the `google-cloud` modules in [1].

[1]: https://github.com/GoogleCloudPlatform/google-cloud-node


DominicKramer:
PTAL

index.js
```diff
@@ -67,6 +67,7 @@ var createLogger = require('./src/logger.js');
  *  Reporting
  */
 
+// TODO: Update this documentation
```
cristiancavalli: Could this documentation be updated before landing?

DominicKramer: I just put the TODO there because there is a separate item for updating jsdoc documentation.

PR #97, opened 2017-01-20
Add koa@next support

cristiancavalli:
- Add interface for koa@next through .koaNext
    method property.
  - Add documentation to readme and new code.
  - Add tests and server scaffolds.
  - Add `npm-install-version` as development
    dependency to use koa@next scaffolds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,21 @@ app.use(function *(){
 app.listen(3000);
 ```
 
+### Using koa@next
+
+```JS
+const errors = require('@google/cloud-errors')();
+const Koa = require('koa');
+const app = new Koa();
+
+// Will attach to 'error' event which includes
+// request/response lifecycle errors
+errors.koaNext(app);
+
+app.use(ctx => ctx.body = 'Hello World!');
+app.listen(3000);
+```
+
 ### Using Restify
 
 ```JS
--- a/tests/test-servers/koaNext_scaffold_server.js
+++ b/tests/test-servers/koaNext_scaffold_server.js
@@ -1,0 +1,46 @@
+/**
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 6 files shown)


cristiancavalli:
@GoogleCloudPlatform/node-team ptal

README.md
```diff
@@ -193,6 +193,21 @@ app.use(function *(){
...
+### Using koa@next
+
+```JS
+const errors = require('@google/cloud-errors')();
+const Koa = require('koa');
+const app = new Koa();
+
+// Will attach to 'error' event which includes
+// request/response lifecycle errors
+errors.koaNext(app);
```
ofrobots: Quick question: I would assume that `koa@next` moves pretty fast. Do we think this integration would be stable, i.e. we won't be broken as `koa@next` changes?

ofrobots: ping

cristiancavalli: I think it would be stable for our purposes and the API surface which we wish to exercise. It may break though given that its OSS. These mocks are meant to break, though, given that the API surface changes so that we can know about it

cristiancavalli: @ofrobots ping

ofrobots: Ack.

tests/test-servers/koaNext_scaffold_server.js
```diff
@@ -0,0 +1,46 @@
+/**
```
ofrobots: Like with the express5, how is this file hooked into the tests?

cristiancavalli:
@ofrobots None of the scaffold servers are hooked into automated test runs like we discussed before

DominicKramer:
Any further work in this area will be done in the [google-cloud-node/error-reporting](https://github.com/GoogleCloudPlatform/google-cloud-node/tree/master/packages/error-reporting) repository and is being tracked by issue [2295](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2295) in that repository.

PR #102, opened 2017-03-28
Fix serviceContext propagation in uncaught reporting

### LINKED ISSUE #101: `serviceContext` not sent with `uncaughtException`
Is it possible that this is due to https://github.com/GoogleCloudPlatform/cloud-errors-nodejs/blob/v0.1.x/lib/interfaces/uncaught.js#L56 calling `sendError` instead of `report` ?

cristiancavalli:
Fixes #101 -- bug where service-context provided during
runtime configuration does not propagate to error messages
created in the uncaughtException behaviour path.

cc @kilianc

### DIFF
--- a/src/interfaces/uncaught.js
+++ b/src/interfaces/uncaught.js
@@ -51,6 +51,8 @@ function handlerSetup(client, config) {
    */
   function uncaughtExceptionHandler(err) {
     var em = new ErrorMessage();
+    em.setServiceContext(config.getServiceContext().service,
+                        config.getServiceContext().version);
     errorHandlerRouter(err, em);
     client.sendError(em, handleProcessExit);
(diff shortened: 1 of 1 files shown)


cristiancavalli:
This is where the end-to-end tests would come into play and that PR is still pending. We do have tests. testing that if these fields are set they are included in the request. In this case we just weren't testing this in relation to uncaughtException event

DominicKramer:
This PR has been closed since the [@google-cloud/error-reporting](https://www.npmjs.com/package/@google-cloud/error-reporting) no longer has this issue since uncaught exceptions must be manually handled.

See that module's [README](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/error-reporting/README.md) for more information about handling uncaught exceptions.

PR #103, opened 2017-03-30
Update README.md

cristiancavalli:
Add specific note stating the side-effects of leaving the `reportUncaughtExceptions` configuration property as `true`.


README.md
```diff
@@ -123,6 +123,18 @@ var errors = require('@google/cloud-errors')({
 });
 ```
 
+#### Notes On Uncaught Exception Reporting
+
+It should be noted that leaving the `reportUncaughtExceptions` configuration option set to its default of `true`
+will cause the error-reporting library to attach to the `process` objects `'uncaughtException'` event which will
+result in the following side effects:
+
+  1. Uncaught exceptions in your script will **not** be printed to `stderr` unless ones application code has itself
```
matthewloring: I think you had written a small example of how this could be done in the original issue. Maybe it could be included here?

ofrobots:
In offline discussions we talked about a doc with a few different options on dealing with this, including a note in the README (here). Is that still the plan?

README.md
cristiancavalli: Good idea -- will include

cristiancavalli:
@ofrobots yes, this was just meant as a stop-gap so that some literature on the subject is put front-and-center

